PR #10, opened 2020-05-11
Remove scale_factor from the attributes of Slide

### LINKED ISSUE #4: should we move scale_factor outside Slide?

alessiamarcolini:
Remove `scale_factor` from the attributes of Slide. 
Closes #4

### DIFF
--- a/src/histolab/slide.py
+++ b/src/histolab/slide.py
@@ -22,15 +22,17 @@ Slide is the main API class for manipulating WSI slides images.
 """
 
 import math
+import os
+import pathlib
+from typing import Tuple, Union
+
+import matplotlib.pyplot as plt
 import ntpath
 import numpy as np
-import openslide
-import os
 import PIL
-import pathlib
-import matplotlib.pyplot as plt
 
-from typing import Tuple, Union
+import openslide
+
 from .util import Time
 
 IMG_EXT = "png"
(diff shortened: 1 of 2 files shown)


src/histolab/slide.py
```diff
@@ -222,19 +236,22 @@ def save_rescaled_slides(self, n=0):
...
+        scale_factor : int, default is 32
+            Image scaling factor
+            
         """
         # TODO: add logger n>total_slide and log thumbnails names
         os.makedirs(self._processed_path, exist_ok=True)
         n = self.total_slides if (n > self.total_slides or n == 0) else n
         for slide in self.slides[:n]:
-            slide.save_thumbnail()
+            slide.save_thumbnail(scale_factor)
```
ernestoarbitrio, line 254: If you change a line you need to write a test, otherwise coveralls will fail cause it consider this a new line missed in tests

src/histolab/slide.py
```diff
@@ -22,15 +22,17 @@
...
 
 import math
-import ntpath
-import numpy as np
-import openslide
 import os
-import PIL
 import pathlib
+from typing import Tuple, Union
+
```
ernestoarbitrio, line 28: why this blank line?

src/histolab/slide.py
```diff
@@ -22,15 +22,17 @@
...
 import matplotlib.pyplot as plt
+import ntpath
+import numpy as np
+import PIL
+
```
ernestoarbitrio, line 33: also here

src/histolab/slide.py
```diff
@@ -120,47 +129,53 @@ def wsi_name(self) -> str:
...
 
-    def _breadcumb(self, directory_path) -> str:
+    def _breadcumb(self, directory_path, scale_factor=32) -> str:
         """Returns a complete path according to the give directory path
 
         Parameters
         ----------
         directory_path: str
+        scale_factor : int, default is 32
+            Image scaling factor
```
ernestoarbitrio, line 139: I would align the param description under the type:
```
scale_factor : int, default is 32
               Image scaling factor
```

src/tests/unit/test_slide.py
```diff
@@ -71,24 +72,24 @@ def it_knows_its_wsi_name(self, wsi_name_fixture):
     def it_knows_its_scaled_image_path(self, scaled_img_path_fixture, resampled_dims_):
         wsi_path, proc_path, wsi_dims, expected_value = scaled_img_path_fixture
         resampled_dims_.return_value = wsi_dims
-        slide = Slide(wsi_path, proc_path, scale_factor=22)
+        slide = Slide(wsi_path, proc_path)
 
-        scaled_img_path = slide.scaled_image_path
+        scaled_img_path = slide.scaled_image_path(22)
```
ernestoarbitrio, line 77: maybe here is better to pass the scale factor in kw mode `scale_factor=22`, i think it's more meaningful and significative for who read the tests

src/tests/unit/test_slide.py
```diff
@@ -138,10 +139,10 @@ def it_knows_its_resampled_array(self, tmpdir, resampled_dims_):
         image = PILImageMock.DIMS_500X500_RGBA_COLOR_155_249_240
         image.save(os.path.join(tmp_path_, "mywsi.png"), "PNG")
         wsi_path = os.path.join(tmp_path_, "mywsi.png")
-        slide = Slide(wsi_path, "processed", 32)
+        slide = Slide(wsi_path, "processed")
         resampled_dims_.return_value = (100, 200, 300, 400)
 
-        resampled_array = slide.resampled_array
+        resampled_array = slide.resampled_array(32)
```
ernestoarbitrio, line 148: same here

src/tests/unit/test_slide.py
```diff
@@ -151,25 +152,27 @@ def it_creates_a_correct_slide_object(self, tmpdir):
...
         with pytest.raises(FileNotFoundError) as err:
-            slide = Slide("wrong/path/fake.wsi", "processed", 32)
+            slide = Slide("wrong/path/fake.wsi", "processed")
             wsi = slide._wsi
 
         assert isinstance(err.value, FileNotFoundError)
         assert str(err.value) == "The wsi path resource doesn't exist"
 
     def or_it_raises_an_PIL_exception(self, tmpdir):
+        print(sys.version_info)
```
ernestoarbitrio, line 173: remove this print statement :D

src/histolab/slide.py
alessiamarcolini: PEP8 suggestions enforced by Black: https://www.python.org/dev/peps/pep-0008/#imports

```
Imports should be grouped in the following order:

1. Standard library imports.
2. Related third party imports.
3. Local application/library specific imports.

You should put a blank line between each group of imports.
```

src/histolab/slide.py
alessiamarcolini: same as before (PEP8)

src/histolab/slide.py
alessiamarcolini: Well, most of the examples of the Numpy style of docstrings simply start the description at one indentation level wrt the parameter name:
* https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard
* https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#module-sphinx.ext.napoleon
* https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html#example-numpy

Otherwise, you'll have the descriptions starting at different columns depending on how long are the parameters name. 

IMHO, having the description at one indentation level for all the parameters is a feature and not a bug :D

ernestoarbitrio: as a matter of legibility, I do prefer the shape I suggest, but ok to me to use the one you prefer ,,, but we need to use that convention in all our module/packages

alessiamarcolini: As another example, docstring generated by `Python Docstring Generator` extension of VS Code: 

image

I would prefer to be consistent with the convention out there

ernestoarbitrio: Ok, I see, you wanna follow third party things instead of your mental pattern (or mine) ... 
OK TO ME  👯‍♀️

PR #11, opened 2020-05-12
rfctr: SlideSet refactoring, unit tests

ernestoarbitrio:
- SlideSet Refactoring
- Unit Tests for the most part of slide.py module


AUTHORS
```diff
@@ -0,0 +1,6 @@
+List of mainly authors
```
alessiamarcolini, line 1: Shouldn't it be "main authors"?

src/histolab/slide.py
```diff
@@ -274,113 +277,105 @@ def slides_dimensions(self):
         ]
 
     @property
-    def slides_stats(self):
-        """Retrieve statistic/graphs of wsi files contained in the dataset"""
-        t = Time()
+    def _wsi_dimensions_list(self) -> List[tuple]:
```
alessiamarcolini: It's not clear to me why we should have both `_wsi_dimensions_list ` and `_slides_dimensions` properties. They represent the same thing (if I'm not missing anything). I would keep `_slides_dimensions` (more structured).

Moreover, we should decide between "slide" or "wsi" naming. They represent the same thing but here it seems not

src/histolab/slide.py
```diff
@@ -274,113 +277,105 @@ def slides_dimensions(self):
...
+        figure, axs = plt.subplots(ncols=2, nrows=2, constrained_layout=True)
+
+        axs[0, 0].scatter(x, y, s=sizes, c=colors, alpha=0.7, cmap="prism")
+        axs[0, 0].set_title("SVS Image Sizes (Labeled with slide name)")
+        axs[0, 0].set_xlabel("width (pixels)")
+        axs[0, 0].set_ylabel("height (pixels)")
+        for i, s in enumerate(self.slides):
+            axs[0, 1].annotate(s.wsi_name, (x[i], y[i]))
+
+        area = [w * h / 1e6 for (w, h) in self._wsi_dimensions_list]
```
alessiamarcolini, line 288: here to change accordingly with the choice above (` _wsi_dimensions_list` vs `_slides_dimensions`)

src/histolab/slide.py
```diff
@@ -245,142 +251,135 @@ def save_thumbnails(self, scale_factor=32, n=0):
...
     @property
-    def _min_size_slide(self):
-        min_size = min(self.slides_dimensions, key=lambda x: x["size"])
-        return {"slide": min_size["wsi"], "height": min_size["size"]}
+    def _min_size_slide(self) -> dict:
+        min_size = min(self._slides_dimensions, key=lambda x: x["size"])
+        return {"slide": min_size["wsi"], "size": min_size["size"]}
+
+    @property
+    def _slides_dimensions(self) -> List[dict]:
```
alessiamarcolini, line 372: It's not clear to me why we should have both `_wsi_dimensions_list ` and `_slides_dimensions` properties. They represent the same thing (if I'm not missing anything). I would keep `_slides_dimensions` (more structured).

Moreover, we should decide between "slide" or "wsi" naming. They represent the same thing but here it seems not

src/histolab/slide.py
ernestoarbitrio: I'm not sure that this is the best choice but slides_dimensions is used for the stats and it's more easy to use to get avg, max etc...
Furthermore wsi_dimensions_list it's just a list of tuple of ints and so more fast than accessing to a dict key each time you need that values. So for now I would keep it separately til we'll find a more pythonic implementation of both

src/histolab/slide.py
ernestoarbitrio: I'm not sure that this is the best choice but slides_dimensions is used for the stats, but IMHO it's more easy to use to get avg, max etc...
Furthermore wsi_dimensions_list it's just a list of tuple of ints and so more fast than accessing to a dict key each time you need that values. So for now I would keep it separately til we'll find a more pythonic implementation of both

AGREE ON NAMING

src/histolab/slide.py
ernestoarbitrio: ???

AUTHORS
ernestoarbitrio: yup, my fault

src/histolab/slide.py
alessiamarcolini: Since we are keeping them both it's ok

PR #23, opened 2020-05-25
Filter test improvement

nicolebussola:
- add integration tests for all image filters functional returning a pil images (tests for "branching" functionals are still missing)
- fix/improve some image filters functionals


tests/integration/test_image_filters.py
```diff
@@ -1,23 +1,31 @@
-from src.histolab.filters.image_filters_functional import invert
-from ..fixtures import RGBA
+import src.histolab.filters.image_filters_functional as imf
```
ernestoarbitrio, line 1: I would refactor imports in this shape:
```
import pytest
import numpy as np

import src.histolab.filters.image_filters_functional as imf

from PIL import ImageChops

from ..fixtures import RGBA, RGB, GS
rom ..util import load_expectation
```

tests/integration/test_image_filters.py
```diff
@@ -28,14 +36,567 @@ def test_invert_filter_with_rgb_hsv_image(pil_image, expectation_image):
...
         np.array(inverted_img), np.array(expected_value)
     )
+
     assert (
         np.unique(np.array(ImageChops.difference(inverted_img, expected_value)))[0] == 0
     )
+
+
+def test_invert_filter_with_gs_image():
+    inverted_img = imf.invert(GS.DIAGNOSTIC_SLIDE_THUMB_GS)
```
ernestoarbitrio, line 54: A test should contain :
- fixtures and assignments
- calls
- assertions

so the tests should be in this format:

def test_invert_filter_with_gs_image():
    img = GS.XXXXX
    expected_value = load_expectaion()

    inverted_img = imf.inverted(img)

    assert A
    assert B
```
```

PR #28, opened 2020-05-26
making histolab pip-installable

ernestoarbitrio:
- setup.py file and all related stuff
- requirements refactoring


setup.py
```diff
@@ -0,0 +1,68 @@
...
+    "scipy==1.4.1",
+]
+
+test_requires = ["pytest", "coverage", "pytest-cov", "coveralls", "pytest-mpl"]
+
+setuptools.setup(
+    name="histolab",
+    version=version,
+    maintainer="Histolab Developers",
+    mantainer_email="ernesto.arbitrio@gmail.com",
```
alessiamarcolini, line 43: typo:
```suggestion
    maintainer_email="ernesto.arbitrio@gmail.com",
```

Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+coverage:
+	pytest --cov-report term-missing --cov=src --cov=tests
+
+dist:
+	rm -rf dist/
+	$(SETUP) sdist bdist_wheel
+
+docs:
+	$(MAKE) -C docs html
```
alessiamarcolini, line 34: Maybe I am missing something, but where is Sphinx configured? I mean, is this enough?

Makefile
```diff
@@ -0,0 +1,45 @@
...
+opendocs:
+	open docs/build/html/index.html
+
+unit-coverage:
+	pytest --cov-report term-missing --cov=src tests/unit
+
+upload:
+	rm -rf dist/
+	$(SETUP) sdist bdist_wheel
+	twine upload dist/*
```
alessiamarcolini, line 45: I don't see twine in any requirements file

Makefile
ernestoarbitrio: it's for future, so i have in mind that we wanna have sphinx and when we wanna release a new doc version we already have the make command to do that

setup.py
ernestoarbitrio: yes it's a typo and BTW i don't wanna use my email, do we wanna create a project discussion list ? eg. histolab-discussions@python.org ? XD

alessiamarcolini: Oh man, you're thinking so big :D 

I think this is a bit overkill for the moment... @nicolebussola what do you think?

Makefile
alessiamarcolini: Yes we want to have sphinx for sure! But having the command here suggests that NOW it's possible to build docs. Maybe we can comment `$(MAKE) -C docs html` and have something like `@echo "Feature not implemented yet"`

setup.py
ernestoarbitrio: as per discussion, but let leave my email for the moment

alessiamarcolini:
LGTM!

PR #29, opened 2020-05-28
Test branching filters

nicolebussola:
- integration tests for image filters
- missing tests for recursive function green_channel_filter TBD


src/histolab/filters/image_filters_functional.py
```diff
@@ -296,6 +296,40 @@ def hysteresis_threshold(
...
+    Returns
+    -------
+    PIL.Image.Image
+        Resulting image where local Otsu threshold values have been
+        applied to original image.
+    """
+    if np.array(img).ndim != 2:
+        raise ValueError("Input must be 2D.")
+    # TODO add function to check real finite number
+    if disk_size is None or disk_size < 0 or disk_size == np.inf:
```
ernestoarbitrio, line 326: I would write a more compact statement using `np.sign`
if `np.sign(disk_size) != 1 or disk_size == np.inf:`

src/histolab/filters/image_filters_functional.py
```diff
@@ -437,10 +445,10 @@ def canny_edges(
     np.ndarray
         Boolean NumPy array representing Canny edge map.
     """
-    if img.size != 2:
+    if len(np.array(img).shape) != 2:
```
ernestoarbitrio, line 448: use `ndim` here instead of len(etc....

src/histolab/util.py
```diff
@@ -38,6 +41,26 @@ def np_to_pil(np_img):
     return Image.fromarray(np_img)
 
 
+def filter_threshold(img: PIL.Image.Image, threshold: float) -> np.ndarray:
```
ernestoarbitrio, line 44: Maybe it's me, but the name `threshold` is odd IMHO. I would expect a number or a range rather than a mask as return value of this util function.

ernestoarbitrio: A more meaningful name could be appreciate

tests/integration/test_image_filters.py
```diff
@@ -633,3 +633,823 @@ def test_hysteresis_threshold_raises_exception_on_thresholds(low, high):
...
+        (
+            GS.DIAGNOSTIC_SLIDE_THUMB_GS,
+            "pil-images-gs/diagnostic-slide-thumb-gs3-local-otsu",
+            0,
+        ),
+        (
+            GS.DIAGNOSTIC_SLIDE_THUMB_GS,
+            "pil-images-gs/diagnostic-slide-thumb-gs4-local-otsu",
+            np.sqrt(2),
+        ),
```
ernestoarbitrio, line 660: For my understanding : what about if the disk size is > than the input image? Is this case meaningful ? So we need a new test case with a big number as disk size?

nicolebussola: not meaningful, but does not raise an Exception

PR #32, opened 2020-06-04
Slide tissue box

nicolebussola:
- add base in test with example images for unit tests
- add unit test for util/threshold_to_mask
- fix some image filters
- change pen filter from "branching functionals" to pil-to-pil functionals


src/histolab/filters/image_filters_functional.py
```diff
@@ -19,6 +19,7 @@
 import math
 from functools import reduce
 
+import operator
```
ernestoarbitrio, line 22: import operator must be under the math import

src/histolab/util.py
```diff
@@ -65,7 +72,55 @@ def threshold_to_mask(img: Image.Image, threshold: float) -> np.ndarray:
...
+    target_size: array_like of int
+        Target (width, height) size of the resulting scaled image
+
+    Returns
+    -------
+    coords: Coordinates
+        Coordinates in the scaled image
+
+    """
+    assert len(reference_size) == 2
```
ernestoarbitrio, line 114: why assert statements here? I think that considering we have tests we can get rid of this lines

tests/unit/test_util.py
```diff
@@ -97,6 +112,236 @@ def test_util_np_to_pil(
...
+            ),
+        ),
+        (IMAGE3_RGBA_BLACK, 2, operator.gt, np.zeros((5, 5, 4), dtype=bool),),
+        (IMAGE3_RGBA_BLACK, 0, operator.lt, np.zeros((5, 5, 4), dtype=bool),),
+        (IMAGE3_RGBA_BLACK, 1, operator.lt, np.ones((5, 5, 4), dtype=bool),),
+        (IMAGE4_RGBA_WHITE, 0, operator.gt, np.ones((5, 5, 4), dtype=bool),),
+        (IMAGE4_RGBA_WHITE, 1, operator.lt, np.zeros((5, 5, 4), dtype=bool),),
+    ),
+)
+def test_util_threshold_to_mask(img, threshold, relate, expected_array):
```
ernestoarbitrio, line 339: 300 lines of fixtures before 3 lines of test if quite strange to read. In these cases I would prefere to have a separate fixture function like in test_slide

src/histolab/util.py
nicolebussola: ok, so should we write a warning or something?

ernestoarbitrio: yes let's manage the exceptions as well as you did in the other modules

nicolebussola: ok, should we write a warning or something?

PR #33, opened 2020-06-06
Conversion filters

alessiamarcolini:
Now is possible to fully compose image and morphological filters using a `Compose` object.

Example:
```python
from PIL import Image
from histolab.filters.morphological_filters import * 
from histolab.filters.image_filters import * 

im = Image.open("image.png")
c = Compose(
        [
            RgbToGrayscale(),
            OtsuThreshold(),
            BinaryDilation(),
            RemoveSmallHoles(),
            BinaryDilation(),
            ApplyMaskImage(im),
        ]
    )
result = c(im)
result.save("result.png")
```

### DIFF
--- a/tests/unitutil.py
+++ b/tests/unitutil.py
@@ -2,8 +2,8 @@
 
 """Functions that make mocking with pytest easier and more readable."""
 
-from unittest.mock import (ANY, PropertyMock, call, create_autospec,  # noqa
-                           patch)
+from unittest.mock import ANY, call  # noqa # isort:skip
+from unittest.mock import create_autospec, patch, PropertyMock  # isort:skip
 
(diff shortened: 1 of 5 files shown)


tests/unitutil.py
```diff
@@ -2,9 +2,11 @@
 
 """Functions that make mocking with pytest easier and more readable."""
 
+from unittest.mock import (ANY, PropertyMock, call, create_autospec,  # noqa
```
ernestoarbitrio, line 5: There was a reason to have that 2 import statement broken in two lines. The `# noqa` comment is used to skip the coverage on ANY and call, but we need to check the coverage on the property mock and patch and so on... so i would leave that two lines as before

PR #38, opened 2020-06-07
Test morph filters

nicolebussola:
-Add unit tests for morphological filter
- Add integration test for morphological filters functional


src/histolab/filters/morphological_filters_functional.py
```diff
@@ -23,7 +23,7 @@
 
 
 def remove_small_objects(
-    np_img: np.ndarray,
+    mask_img: np.ndarray,
```
ernestoarbitrio, line 26: to be compliant with the other names like `np_mask` i would use the same name here

PR #40, opened 2020-06-08
Add tile module + BinaryFillHoles filter

alessiamarcolini:
* Add tile module
* Add BinaryFillHoles morphological filter, needed to check if the tile has enough tissue.

*Note for reviewers: it may be convenient to look only at the final diff instead of the diff for single commits.*


src/histolab/tile.py
```diff
@@ -0,0 +1,161 @@
...
+            Whether the image has enough tissue, i.e. if the proportion of tissue
+            over the total area of the image is more than ``threshold`` and the image
+            variance after morphological operations is more than
+            ``near_zero_var_threshold``.
+        """
+
+        if self._is_almost_white:
+            return False
+
+        if self._is_all_tissue_or_no_tissue(near_zero_var_threshold):
```
nicolebussola, line 57: I do not remember why "all tissue" is a problem

alessiamarcolini: It's not actually a problem. The point is: if we check for "no tissue" using the variance of the image, "no tissue" is the same as "all tissue". We probably need to find another way to determine if we have no tissue in the tile, then.

src/histolab/tile.py
```diff
@@ -0,0 +1,162 @@
...
+from .util import lazyproperty
+
+
+class Tile:
+    def __init__(self, image: Image.Image, coords: CoordinatePair, level: int = 0):
+        self._image = image
+        self._level = level
+        self._coords = coords
+
+    @lazyproperty
```
ernestoarbitrio, line 20: why this property to return an attribute of the class? We do already have self._image without this prop

src/histolab/tile.py
```diff
@@ -0,0 +1,162 @@
...
+    def image(self) -> Image.Image:
+        return self._image
+
+    @lazyproperty
+    def level(self) -> int:
```
ernestoarbitrio, line 25: also here

src/histolab/tile.py
```diff
@@ -0,0 +1,162 @@
...
+        return self._level
+
+    @lazyproperty
+    def coords(self) -> CoordinatePair:
```
ernestoarbitrio, line 29: and here

src/histolab/tile.py
ernestoarbitrio: IMHO this prop name is odd, considering that it returns a bool, can we call it just `_is_all_tissue` ?

src/histolab/tile.py
```diff
@@ -0,0 +1,162 @@
...
+
+        if self._is_all_tissue_or_no_tissue(near_zero_var_threshold):
+            return False
+
+        if not self._has_tissue_more_than_percent(tissue_percent):
+            return False
+
+        return True
+
+    def save(self, path):
```
ernestoarbitrio, line 66: python typing missed

src/histolab/tile.py
```diff
@@ -0,0 +1,162 @@
...
+from PIL import Image
+
+from .filters import image_filters as imf
+from .filters import morphological_filters as mof
+from .filters.util import mask_percent
+from .types import CoordinatePair
+from .util import lazyproperty
+
+
+class Tile:
```
ernestoarbitrio, line 14: where's the docstring? XD

src/histolab/tile.py
alessiamarcolini: The point is that this evaluates to True both for cases of ALL tissue and cases of NO tissue at all. Mmmmm

src/histolab/tile.py
alessiamarcolini: Yes but is a way to expose `image` as a read-only attribute, right?

src/histolab/tile.py
alessiamarcolini: totally missed :D

src/histolab/tile.py
ernestoarbitrio: yes but it's a bool and the name can be more simple than that, can we try to change that name? @nicolebussola what do you think about it?

src/histolab/tile.py
ernestoarbitrio: oh ok i didn't realize we wanna have that as a read-only attr

alessiamarcolini: yeah, Tile is a read-only container to me

PR #41, opened 2020-06-08
Test biggest tissue box mask and connected props

ernestoarbitrio:
- Unit tests for `biggest_tissue_box_mask`
- Unit tests for all related property and functions

NOTE: integration tests for `biggest_tissue_box_mask` will be in another PR


src/histolab/slide.py
```diff
@@ -173,8 +173,7 @@ def thumbnail_path(self) -> str:
 
     # ---private interface methods and properties---
 
-    @staticmethod
```
alessiamarcolini, line 176: Why removing `@staticmethod`? The method is not accessing the instance nor the class, and it's independent from everything else around it.

src/histolab/slide.py
```diff
@@ -247,8 +246,7 @@ def _main_tissue_areas_mask_filters(self) -> imf.Compose:
         )
         return filters
 
-    @staticmethod
```
alessiamarcolini, line 250: same as above

src/histolab/slide.py
```diff
@@ -264,8 +262,7 @@ def _region_coordinates(region: Region) -> CoordinatePair:
         y_ul, x_ul, y_br, x_br = region.bbox
         return CoordinatePair(x_ul, y_ul, x_br, y_br)
 
-    @staticmethod
```
alessiamarcolini, line 267: same as above

tests/unit/test_image_filters.py
```diff
@@ -103,7 +103,7 @@ def it_calls_adaptive_equalization_functional(self, request):
     def it_calls_local_equalization_functional(self, request):
         image = PILImageMock.DIMS_500X500_RGBA_COLOR_155_249_240
         F_local_equalization = function_mock(
-            request, "src.histolab.filters.image_filters_functional.local_equalization",
```
alessiamarcolini, line 106: I think it's Black which is adding the trailing comma here. What are your settings?

src/histolab/slide.py
ernestoarbitrio: Sure thing, but unittes+mcok has a known issue when we use staticmethods decorators, when we figure out how to fix this kind of problems we can re-decorate that methods

tests/unit/test_image_filters.py
ernestoarbitrio: the one in our project.toml:
```
[tool.black]
line-length = 88
include = '\.pyi?$'
exclude = '''
/(
    \.git
  | \.hg
  | \.mypy_cache
  | \.tox
  | \.venv
  | _build
  | buck-out
  | build
  | dist
)/
'''
```

src/histolab/slide.py
alessiamarcolini: Didn't know about that!

PR #47, opened 2020-06-11
Data module for loading tissue samples

ernestoarbitrio:
based on https://scikit-image.org/docs/stable/api/skimage.data.html this new module allows users to use sample data from local data dir within the repo or using external samples.

When samples are not available in the data module they're downloaded from the registry url using pooch


src/histolab/data/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+    Path of the local file as a python string.
+    Raises
+    ------
+    KeyError:
+        If the filename is not known to the histolab distribution.
+    ModuleNotFoundError:
+        If the filename is known to the histolab distribution but pooch
+        is not installed.
+    ConnectionError:
+        If histolab is unable to connect to the internet but the
```
alessiamarcolini, line 127: I'd say:

If the dataset has not been downloaded yet and histolab is unable to connect to the internet

src/histolab/data/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+        raise ModuleNotFoundError(
+            "The requested file is part of the histolab distribution, "
+            "but requires the installation of an optional dependency, pooch. "
+            "To install pooch, use your preferred python package manager. "
+            "Follow installation instruction found at "
+            "https://www.fatiando.org/pooch/latest/install.html"
+        )
+
+    # Case 4:
+    # Pooch needs to download the data. Let the image fetcher to search for
```
alessiamarcolini, line 164: remove "to" before search

src/histolab/data/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+    for filename in legacy_registry:
+        _fetch(filename)
+
+
+if has_pooch:
+    _init_pooch()
+
+
+def _load_svs(f):
+    """Load an image file located in the data directory."""
```
alessiamarcolini, line 194: missing docstring for Parameters, Returns and Raises

src/histolab/data/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+    https://api.gdc.cancer.gov/data/9c960533-2e58-4e54-97b2-8454dfb4b8c8
+
+    Access: open
+
+    Returns
+    -------
+    breast_tissue : OpenSlide object
+        Breast tissue Whole-Slide-Image
+    """
+    return _load_svs("tcga/breast/9c960533-2e58-4e54-97b2-8454dfb4b8c8")
```
alessiamarcolini, line 264: no extension here?

src/histolab/data/_registry.py
```diff
@@ -0,0 +1,27 @@
...
+legacy_datasets = ["cmu_small_region.svs"]
+
+# Registry of datafiles that can be downloaded along with their SHA256 hashes
+# To generate the SHA256 hash, use the command
+# openssl sha256 filename
+registry = {
+    "data/cmu_small_region.svs": "ed92d5a9f2e86df67640d6f92ce3e231419ce127131697fbbce42ad5e002c8a7",
+    "aperio/JP2K-33003-1.svs": "6205ccf75a8fa6c32df7c5c04b7377398971a490fb6b320d50d91f7ba6a0e6fd",
+    "aperio/JP2K-33003-2.svs": "1a13cef86b55b51127cebd94a1f6069f7de494c98e3e708640d1ce7181d9e3fd",
+    "tcga/breast/9c960533-2e58-4e54-97b2-8454dfb4b8c8": "03f542afa2d70224d594b2cca33b99977a5c0e41b1a8d03471ab3cf62ea3c4b3",
```
alessiamarcolini, line 13: maybe we could use the TCGA "project name" such as TCGA-BRCA instead of the general "breast"... being more specific could help the scientific community

src/histolab/data/util.py
```diff
@@ -0,0 +1,56 @@
...
+import tempfile
+import urllib.parse
+import urllib.request
+from contextlib import contextmanager
+from urllib.error import HTTPError, URLError
+
+URL_REGEX = re.compile(r"http://|https://|ftp://|file://|file:\\")
+
+
+def _is_url(filename):
```
alessiamarcolini, line 30: type annotation missing and docstring too less detailed

src/histolab/data/util.py
```diff
@@ -0,0 +1,56 @@
...
+    """Return True if string is an http or ftp path."""
+    return isinstance(filename, str) and URL_REGEX.match(filename) is not None
```
alessiamarcolini, line 32: "is not None" -> not pythonic

src/histolab/data/util.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@contextmanager
+def file_or_url_context(resource_name):
```
alessiamarcolini, line 36: type annotations missing and docstring too less detailed

src/histolab/data/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+
+from .. import __version__
+from ._registry import legacy_registry, registry, registry_urls
+
+legacy_data_dir = os.path.abspath(os.path.dirname(__file__))
+histolab_distribution_dir = os.path.join(legacy_data_dir, "..")
+
+try:
+    from pooch.utils import file_hash
+except ModuleNotFoundError:
```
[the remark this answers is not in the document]
ernestoarbitrio: nope cause we match hashes for files even if wihtout pooch

src/histolab/data/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+    # Function taken from
+    # https://github.com/fatiando/pooch/blob/master/pooch/utils.py
+    def file_hash(fname, alg="sha256"):
+        """
+        Calculate the hash of a given file.
```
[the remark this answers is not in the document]
ernestoarbitrio: u right

src/histolab/data/__init__.py
ernestoarbitrio: npe this is the trick you sent me for tcga ... so no extension here to be compliant with the API get request

alessiamarcolini: well... so here the UUID becomes the filename? :( 
the point is that for TCGA the filename has a specific meaning (see https://docs.gdc.cancer.gov/Encyclopedia/pages/TCGA_Barcode/)

src/histolab/data/_registry.py
ernestoarbitrio: this is one of the question i'm gonna leave you both @alessiamarcolini and @nicolebussola. This file is just a registry the users will get the svs using the function name in the init of this module. So the important thing is that the meaningful name should be in there, isn't it?

src/histolab/data/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+        hasher = hashlib.new(alg)
+        with open(fname, "rb") as fin:
+            buff = fin.read(chunksize)
+            while buff:
+                hasher.update(buff)
+                buff = fin.read(chunksize)
+        return hasher.hexdigest()
+
+
+def create_image_fetcher():
```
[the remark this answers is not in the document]
ernestoarbitrio: this function is now private and i can't use type annotations here because pooch is imported in the try except statement. So the return values are pooch.Pooch obj + str, but pooch isn't imported by default so .... i would leave it as is

PR #51, opened 2020-06-14
Add Tiler module with RandomTiler

alessiamarcolini:
At the moment the extraction at level 0 is really slow (due to the size of the tissue mask I suppose, maybe we can think to optimize somehow later).

### DIFF
--- a/src/histolab/slide.py
+++ b/src/histolab/slide.py
@@ -90,13 +90,12 @@ class Slide(object):
 
     @lazyproperty
     def biggest_tissue_box_mask(self) -> np.ndarray:
-        """Returns the coordinates of the box containing the max area of tissue.
+        """Returns the binary mask of the box containing the max area of tissue.
 
         Returns
         -------
-        box_coords: Coordinates
-            [x_ul, y_ul, x_br, y_br] coordinates of the box containing the
-            max area of tissue.
+        mask: np.ndarray
+            Binary mask of the box containing the max area of tissue.
 
         """
 
@@ -105,7 +104,7 @@ class Slide(object):
 
         thumb_mask = filters(thumb)
         regions = self._regions_from_binary_mask(thumb_mask)
-        biggest_region = self._biggest_regions(regions, n=1)
+        biggest_region = self._biggest_regions(regions, n=1)[0]
         biggest_region_coordinates = self._region_coordinates(biggest_region)
         thumb_bbox_mask = polygon_to_mask_array(
             (1000, 1000), biggest_region_coordinates
--- a/src/histolab/slide.py
+++ b/src/histolab/slide.py
@@ -134,12 +134,7 @@ class Slide(object):
             Image containing the selected tile
         """
 
-        if (
-            coords.x_ul >= self.dimensions[0]
-            or coords.x_br >= self.dimensions[0]
-            or coords.y_ul >= self.dimensions[1]
-            or coords.y_br >= self.dimensions[1]
-        ):
+        if not self._are_valid_coords(coords):
             # OpenSlide doesn't complain if the coordinates for extraction are wrong,
             # but it returns an odd image.
             raise ValueError(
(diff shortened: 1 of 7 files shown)


src/histolab/tiler.py
```diff
@@ -77,20 +77,48 @@ def __init__(
...
         self.prefix = prefix
         self.suffix = suffix
 
+    @property
+    def tile_size(self) -> Tuple[int, int]:
+        return self._valid_tile_size
+
+    @tile_size.setter
+    def tile_size(self, tile_size_: Tuple[int, int]):
+        if tile_size_[0] < 1 or tile_size_[1] < 1:
```
ernestoarbitrio, line 95: note: 0 is considered positive, so or this condition is wrong or the error message is flaky :D

tests/unit/test_tiler.py
```diff
@@ -0,0 +1,61 @@
...
+        image = PILImageMock.DIMS_50X50_RGB_RANDOM_COLOR
+        image.save(os.path.join(tmp_path_, "mywsi.png"), "PNG")
+        slide_path = os.path.join(tmp_path_, "mywsi.png")
+        slide = Slide(slide_path, "processed")
+        random_tiler = RandomTiler((512, 512), 10, 3)
+
+        with pytest.raises(IndexError) as err:
+            random_tiler.extract(slide)
+
+        assert isinstance(err.value, IndexError)
```
ernestoarbitrio, line 39: I'm not sure this is a significative test, for sure here we're simulating a wrong user behaviour ... but in future these kind of exceptions could be managed as custom, so we will have "our" exception rather than checking for the builtin exception. NOTE: this kind of error returned to a standard user is not meaningful

tests/unit/test_tiler.py
```diff
@@ -0,0 +1,61 @@
...
+            RandomTiler((512, 512), 10, 0, max_iter=3)
+
+        assert isinstance(err.value, ValueError)
+        assert (
+            str(err.value)
+            == "The maximum number of iterations (3) must be grater than or equal to the maximum number of tiles (10)."
+        )
+
+    @pytest.fixture
+    def Slide_(self, request):
```
ernestoarbitrio, line 60: Get rid of this fixture, it's no longer used i believe

src/histolab/tiler.py
```diff
@@ -0,0 +1,278 @@
...
+            raise ValueError(f"Tile size cannot be negative ({tile_size_})")
+        self._valid_tile_size = tile_size_
+
+    @property
+    def level(self) -> int:
```
ernestoarbitrio, line 100: docstring missing

src/histolab/tiler.py
```diff
@@ -0,0 +1,278 @@
...
+                yield tile, tile_wsi_coords
+                valid_tile_counter += 1
+
+            if self.max_iter and iteration > self.max_iter:
+                break
+
+            if valid_tile_counter > self.n_tiles:
+                break
+
+    def _tile_filename(self, tile_wsi_coords, tiles_counter):
```
ernestoarbitrio, line 267: docstring missing

tests/unit/test_tiler.py
ernestoarbitrio: `or_it_has_not_available_level_value`

alessiamarcolini: Yes, we are simulating a wrong user behavior that could totally happen.

To have a more meaningful message I could catch the exception in `level_dimensions()` and raise a `ValueError` instead - later to be turned into our custom exception. What do you think?

ernestoarbitrio: i do agree let's do this thing :D

src/histolab/slide.py
```diff
@@ -98,13 +111,52 @@ def biggest_tissue_box_mask(self):
...
         )
         return resize_mask(thumb_bbox_mask, self.dimensions)
 
+    def extract_tile(self, coords: CoordinatePair, level: int) -> Tile:
+        """Extract a tile of the image at the selected level.
+
+        Parameters
+        ----------
+        coords : CoordinatePair
+            Coordinates in the first level (0)
```
nicolebussola, line 127: I would change in "coordinates at level 0"

PR #54, opened 2020-06-15
Add show method for displaying its thumbnail

ernestoarbitrio:
New method `show` in `Slide` that displays the slide thumbnail if available

### DIFF
--- a/src/histolab/slide.py
+++ b/src/histolab/slide.py
@@ -173,8 +173,8 @@ class Slide(object):
         try:
             thumbnail = PIL.Image.open(self.thumbnail_path)
             thumbnail.show()
-        except Exception as e:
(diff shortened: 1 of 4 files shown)


src/histolab/slide.py
```diff
@@ -165,6 +165,17 @@ def scaled_image_path(self, scale_factor: int = 32) -> str:
...
 
+    def show(self) -> None:
+        """Display the slide thumbnail.
+
+        NOTE: A new window of your OS image viewer will be opened.
+        """
+        try:
+            thumbnail = PIL.Image.open(self.thumbnail_path)
+            thumbnail.show()
+        except Exception as e:
```
alessiamarcolini, line 176: could you restrict to some more specific exception?

PR #69, opened 2020-06-17
Use sparse arrays for box masks instead of regular np arrays

alessiamarcolini:
This greatly improves performance tissue box retrieval and tiles extraction


alessiamarcolini:
Fix #70

PR #77, opened 2020-06-18
Fix docs (type annotations and docstrings)

alessiamarcolini:


src/histolab/filters/image_filters_functional.py
```diff
@@ -21,60 +21,61 @@
 from functools import reduce
 
 import numpy as np
+import PIL
+import PIL.ImageOps
```
nicolebussola, line 25: Is it necessary to import also PIL.ImageOps?

src/histolab/util.py
```diff
@@ -23,15 +23,16 @@
 from typing import Tuple
 
 import numpy as np
+import PIL
+import PIL.ImageDraw
```
nicolebussola, line 27: same question here

src/histolab/filters/image_filters_functional.py
alessiamarcolini: it seems necessary. Importing only PIL and calling `PIL.ImageOps.function` seems not possible. @ernestoarbitrio any clues?

alessiamarcolini: Offline he said he has no clues. I'll leave as it is for the moment

PR #109, opened 2020-06-23
Add Grid tiler

### LINKED ISSUE #99: Add Grid Tiler
We need a Tiler to extract the tiles from a "grid". Tiles may be overlapping or not, depending on user choice.

alessiamarcolini:
Fix #99

### DIFF
--- a/src/histolab/tiler.py
+++ b/src/histolab/tiler.py
@@ -117,18 +117,26 @@ class GridTiler(Tiler):
         for region in regions:  # at the moment there is only one region
             bbox_coordinates = region_coordinates(region)
 
-            n_tiles_row = (bbox_coordinates.x_br - bbox_coordinates.x_ul) // tile_w_lvl
-            n_tiles_column = (
-                bbox_coordinates.y_br - bbox_coordinates.y_ul
-            ) // tile_h_lvl
+            n_tiles_row = (bbox_coordinates.x_br - bbox_coordinates.x_ul) // (
+                tile_w_lvl - self.pixel_overlap
+            )
+            n_tiles_column = (bbox_coordinates.y_br - bbox_coordinates.y_ul) // (
+                tile_h_lvl - self.pixel_overlap
+            )
 
             x_ul_lvl_offset = bbox_coordinates.x_ul
             y_ul_lvl_offset = bbox_coordinates.y_ul
 
             for i in range(n_tiles_row):
                 for j in range(n_tiles_column):
-                    x_ul_lvl = x_ul_lvl_offset + tile_w_lvl * j
-                    y_ul_lvl = y_ul_lvl_offset + tile_h_lvl * i
+                    x_ul_lvl = x_ul_lvl_offset + tile_w_lvl * j - self.pixel_overlap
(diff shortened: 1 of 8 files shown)


src/histolab/tiler.py
```diff
@@ -1,21 +1,250 @@
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from typing import Tuple
 
 import numpy as np
 import sparse
+from typing_extensions import Protocol, runtime_checkable
```
ernestoarbitrio, line 6: considering that 3.7 already have this 2 module in the typing std lib ... here I would use a try except statement to import them from the standard lib and if exception is raised let's import from typing_extension

tests/unit/test_tiler.py
```diff
@@ -323,3 +352,283 @@ def tile_filename_fixture(self, request):
...
+            ),
+            (
+                CoordinatePair(0, 10, 0, 10),
+                CoordinatePair(0, 10, 0, 10),
+                True,
+                [True, False],
+                1,
+            ),
+        ),
+    )
```
ernestoarbitrio, line 527: I would avoid using 100 lines fixture in the parametrize decorator, I do prefer to have a separate fixture method to be passed as arg in the test function

src/histolab/tiler.py
```diff
@@ -1,21 +1,250 @@
...
+
+        return tile_filename
+
     @abstractmethod
     def extract(self, slide: Slide):
         raise NotImplementedError
 
 
+class GridTiler(Tiler):
+    def __init__(
```
nicolebussola, line 105: docstring?

src/histolab/tiler.py
```diff
@@ -1,21 +1,250 @@
...
+                    y_br_lvl = y_ul_lvl + tile_h_lvl
+
+                    tile_wsi_coords = scale_coordinates(
+                        reference_coords=CoordinatePair(
+                            x_ul_lvl, y_ul_lvl, x_br_lvl, y_br_lvl
+                        ),
+                        reference_size=slide.level_dimensions(level=self.level),
+                        target_size=slide.level_dimensions(level=0),
+                    )
+                    yield tile_wsi_coords
```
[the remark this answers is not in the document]
alessiamarcolini: working on it :D

src/histolab/tiler.py
```diff
@@ -7,21 +7,317 @@
...
+            Iterator of tiles' CoordinatePair
+        """
+        box_mask_lvl = self.box_mask_lvl(slide)
+
+        regions = regions_from_binary_mask(box_mask_lvl.todense())
+        for region in regions:  # at the moment there is only one region
+            bbox_coordinates = region_coordinates(region)
+            yield from self._grid_coordinates_from_bbox_coordinates(
+                bbox_coordinates, slide
+            )
```
alessiamarcolini, line 252: @ernestoarbitrio any better? if you have any suggestions, will be more than welcome

ernestoarbitrio: not really ... let's move on for now

PR #124, opened 2020-07-30
Add Lambda filter

### LINKED ISSUE #119: Lambda filter
We should have a Lambda filter Class to enable filter extension with custom functions. 

Inspired by https://pytorch.org/docs/stable/torchvision/transforms.html#torchvision.transforms.Lambda

alessiamarcolini:
Fix #119

### DIFF
--- a/src/histolab/filters/image_filters.py
+++ b/src/histolab/filters/image_filters.py
@@ -16,7 +16,7 @@
 # limitations under the License.
 # ------------------------------------------------------------------------
 
-from typing import Any, List, Union
+from typing import Any, Callable, List, Union
 
 import numpy as np
 import PIL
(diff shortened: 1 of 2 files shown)


src/histolab/filters/image_filters.py
```diff
@@ -43,6 +43,34 @@ def __call__(self, img: PIL.Image.Image) -> Union[PIL.Image.Image, np.ndarray]:
...
+
+    Returns
+    -------
+    PIL.Image.Image
+        The image with the function applied.
+    """
+
+    def __init__(self, lambd: Callable[[PIL.Image.Image], PIL.Image.Image]) -> None:
+        assert callable(lambd), repr(type(lambd).__name__) + " object is not callable"
+        self.lambd = lambd
```
nicolebussola, line 65: maybe lmb istead of lambd?

alessiamarcolini: mmm I just copied it from torchvision, do you think lmb is better..?

nicolebussola: uh ok.... I just do not like lambda without the a.... but ok let's use torchvision style

alessiamarcolini: It's not possible to use lambda because it's a python keyword 🤷🏻‍♀️

PR #126, opened 2020-07-30
Save tiles in slide's processed_path

### LINKED ISSUE #114: Tiles generated from a Tiler are not saved inside the "processed_path" of the Slide
By default, tiles generated from the tilers are not saved inside the `processed_path` of the Slide. We are now using the `prefix` attribute of the tiler to point to the same `processed_path`, when it should be used to specify a prefix for the filename or a possible subfolder

alessiamarcolini:
Fix #114

### DIFF
--- a/src/histolab/tiler.py
+++ b/src/histolab/tiler.py
@@ -1,3 +1,4 @@
+import os
 from abc import abstractmethod
 from typing import Tuple
 
@@ -178,7 +179,9 @@ class GridTiler(Tiler):
 
         for tiles_counter, (tile, tile_wsi_coords) in enumerate(grid_tiles):
             tile_filename = self._tile_filename(tile_wsi_coords, tiles_counter)
-            tile.save(tile_filename)
+            root_folder = slide.processed_path
+            full_tile_path = os.path.join(root_folder, tile_filename)
+            tile.save(full_tile_path)
             print(f"\t Tile {tiles_counter} saved: {tile_filename}")
 
         print(f"{tiles_counter+1} Grid Tiles have been saved.")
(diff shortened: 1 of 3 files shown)


src/histolab/tiler.py
```diff
@@ -410,7 +413,9 @@ def extract(self, slide: Slide):
         tiles_counter = 0
         for tiles_counter, (tile, tile_wsi_coords) in enumerate(random_tiles):
             tile_filename = self._tile_filename(tile_wsi_coords, tiles_counter)
-            tile.save(tile_filename)
+            root_folder = slide.processed_path
```
ernestoarbitrio, line 416: maybe this assignment is not mandatory
`os.path.join(self.processed_path, tile_filename)` is enough i believe

tests/unit/test_tiler.py
```diff
@@ -297,6 +297,34 @@ def it_knows_its_box_mask_lvl(
...
+
+        random_tiler.extract(slide)
+
+        assert os.path.exists(
+            os.path.join(tmp_path_, "processed", "tile_0_level2_0-10-0-10.png")
+        )
+        assert os.path.exists(
+            os.path.join(tmp_path_, "processed", "tile_1_level2_0-10-0-10.png")
+        )
+
```
ernestoarbitrio, line 327: can you add some assertion on the called method? Which args or how many time the method or the property are called? Just to be sure that the function is formally correct. IMHO the only assertions on the path are not completely meaninful.

tests/unit/test_tiler.py
```diff
@@ -541,6 +569,32 @@ def and_doesnt_raise_error_with_wrong_coordinates(self, request, tmpdir):
...
+        grid_tiler.extract(slide)
+
+        assert os.path.exists(
+            os.path.join(tmp_path_, "processed", "tile_0_level2_0-10-0-10.png")
+        )
+        assert os.path.exists(
+            os.path.join(tmp_path_, "processed", "tile_1_level2_0-10-0-10.png")
+        )
+
```
ernestoarbitrio, line 597: also here

alessiamarcolini:
The tiles should be saved in a subfolder of slide's processed path called "tiles"

PR #127, opened 2020-08-03
Travis build on macOS and Windows

alessiamarcolini:
There are failing tests on Windows, needed to be fixed later

EDIT: all the tests are passing now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,13 @@ jobs:
       language: shell       # 'language: python' is an error on Travis CI macOS
       before_install:
         - brew update
+        - brew install openslide
     - name: "Python 3.7.4 on macOS"
       os: osx
       osx_image: xcode11.2  # Python 3.7.4 running on macOS 10.14.4
       language: shell       # 'language: python' is an error on Travis CI macOS
+      before_install:
+        - brew install openslide
 
 # command to install dependencies
 install:
@@ -44,7 +47,5 @@ notifications:
   slack: mpba:fBrjUWlUSa3XKeCA8zFsTFu2
 
 addons:
-  homebrew:
-    packages: openslide
   apt:
     packages: openslide-tools
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,18 +15,16 @@ jobs:
       before_install:
         - brew update
         - brew install openslide
-        - brew unlink python # ONLY if you have installed (with brew) another version of python 3
-        - brew install --ignore-dependencies https://raw.githubusercontent.com/Homebrew/homebrew-core/f2a764ef944b1080be64bd88dca9a1d80130c558/Formula/python.rb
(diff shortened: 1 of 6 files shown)


alessiamarcolini:
I just noticed that macOS uses the brew's Python (3.8) instead of the system's one

tests/unit/test_slide.py
```diff
@@ -694,7 +701,14 @@ def it_knows_its_slides(self, tmpdir):
...
         assert isinstance(err.value, FileNotFoundError)
-        assert str(err.value) == "[Errno 2] No such file or directory: 'fake/path'"
+
+        if os.environ.get("TRAVIS_OS_NAME", "linux") == "windows":
+            assert (
+                str(err.value)
+                == "[WinError 3] The system cannot find the path specified: 'fake/path'"
+            )
+        else:
+            assert str(err.value) == "[Errno 2] No such file or directory: 'fake/path'"
```
ernestoarbitrio, line 711: I think that this statement should be replaced with something more general. For example the `errno` attribute on the error should be the same on all platforms. You will get WindowsError exceptions on Windows, but since this is a subclass of OSError the same "except OSError:" block will catch it. Windows does have its own error codes, and these are accessible as .winerror, but the .errno attribute should still be present, and usable in a cross-platform way.

Something like:
`OSError(errno.EEXIST, 'Error Message here')`

.travis.yml
```diff
@@ -1,18 +1,53 @@
...
+      language: shell       # 'language: python' is an error on Travis CI Windows
+      before_install:
+        - choco install python --version 3.6.8
+        - python -m pip install --upgrade pip
+        - wget https://github.com/openslide/openslide-winbuild/releases/download/v20171122/openslide-win64-20171122.zip -P /c/downloads
+        - 7z e /c/downloads/openslide-win64-20171122.zip -aoa
+      env: PATH=/c/Python36:/c/Python36/Scripts:/c/downloads/openslide-win64-20171122/bin:$PATH
+    - name: "Python 3.7 on Windows"
+      os: windows           # Windows 10.0.17134 N/A Build 17134
+      language: shell       # 'language: python' is an error on Travis CI Windows
```
ernestoarbitrio, line 39: rlly? windows sucks

tests/unit/test_slide.py
alessiamarcolini: Unfortunately, there is no `errno` attribute in `err`. Moreover, the error message is decided by the OS, it is not ours, so I cannot even try something like `"error message" in err.value`. Any suggestions?

ernestoarbitrio: `errno` is in the std lib
so my huntch is 
```
import errno
assert str(err.value) == errno.EEXISTS
```

BTW here the reference https://python.readthedocs.io/en/latest/library/errno.html

ernestoarbitrio (approved):

PR #128, opened 2020-08-04
Instead of using the tissue box mask at level 0, keep it at thumbnail dimensions + drop sparse dependency

alessiamarcolini:
To fix a critical memory issue, instead of calculating the tissue box mask at level 0 (and then scaling it to the extraction level), we can keep it at thumbnail dimensions and then scale accordingly all the subsequent operations.

We are dropping also the `sparse` dependency (used to store the mask sparse array)


alessiamarcolini:
The coverage report says:

`src/histolab/exceptions.py                                     9      1      4      1         84.62%        5->8, 8`

How to test lines from 5 to 8? @ernestoarbitrio any advice?

src/histolab/tiler.py
```diff
@@ -28,8 +28,8 @@ class Tiler(Protocol):
     tile_size: int
 
     @lru_cache(maxsize=100)
-    def box_mask(self, slide: Slide) -> sparse._coo.core.COO:
-        """Return binary mask at level 0 of the box to consider for tiles extraction.
+    def box_mask_thumb(self, slide: Slide) -> np.ndarray:
+        """Return binary mask at thumbnail dimensions of the box for tiles extraction.
 
         The mask pixels set to True will be the ones corresponding to the tissue box.
```
nicolebussola, line 34: "will be the ones corresponding to" -> "correspond to"

src/histolab/tiler.py
```diff
@@ -28,8 +28,8 @@ class Tiler(Protocol):
     tile_size: int
 
     @lru_cache(maxsize=100)
-    def box_mask(self, slide: Slide) -> sparse._coo.core.COO:
-        """Return binary mask at level 0 of the box to consider for tiles extraction.
+    def box_mask_thumb(self, slide: Slide) -> np.ndarray:
+        """Return binary mask at thumbnail dimensions of the box for tiles extraction.
```
nicolebussola, line 32: ""Return binary mask at thumbnail dimensions of the box for tiles extraction." --> "Return binary mask, at thumbnail level, of the box for tiles extraction.

src/histolab/tiler.py
```diff
@@ -40,38 +40,12 @@ def box_mask(self, slide: Slide) -> sparse._coo.core.COO:
 
         Returns
         -------
-        sparse._coo.core.COO
-            Extraction mask at level 0
+        np.ndarray
+            Extraction mask at thumbnail dimensions
```
nicolebussola, line 44: thumbnail level

PR #129, opened 2020-08-05
Add ScoreTiler

alessiamarcolini:
A `ScoreTiler` is basically a `GridTiler` (because it extracts all the tiles following a grid structure), but it saves only the first `n_tiles` with the highest score, according to a scoring function. If `n_tiles` is equal to 0, it saves all the tiles (exact same behavior as a `GridTiler`)

### Scorer
The scoring functions are implemented in the `scorer` module. All the scorers are classes with a `__call__` method, which takes as parameter a `Tile` object and returns a float. These characteristics are implemented in the `Scorer` protocol.  At the moment there is only a `RandomScorer`, which returns a random number between 0 and 1, not considering the tile, but it serves as an initial example.

#### Example
```python
from histolab.scorer import RandomScorer

random_scorer = RandomScorer()
tile = Tile(img, coords, level)
score = random_scorer(tile)
```

### ScoreTiler
The `ScoreTiler` (other than the parameters required by a `GridTiler`) needs a scoring function (which implements the `Scoring` protocol).

#### Example
```python
# following from the previous example

from histolab.tiler import ScoreTiler
from histolab.slide import Slide

slide = Slide("path/to/slide", "path/to/processed")
score_tiler = ScoreTiler(random_scorer, tile_size=(512,512), n_tiles=2, level=0, ...)
score_tiler.extract(slide)
```


alessiamarcolini:
We should save a .csv file containing the filenames of the tiles and the associated scores

src/histolab/tiler.py
```diff
@@ -1,17 +1,18 @@
 from abc import abstractmethod
-from typing import Tuple
+from typing import List, Tuple
 
 import numpy as np
-import sparse
 
+from histolab.exceptions import LevelError
```
ernestoarbitrio, line 6: I do have a concern about this import. I'd prefer relative imports also because we could have some issues with the coverage report.

tests/unit/test_exceptions.py
```diff
@@ -0,0 +1,56 @@
...
+
+        if type(args) == list:
+            level_error = exp.LevelError(*args)
+        else:
+            level_error = exp.LevelError(args)
+
+        if type(args) == list:
+            _init.assert_called_once_with(ANY, *args)
+        else:
+            _init.assert_called_once_with(ANY, args)
```
ernestoarbitrio, line 21: why 2 equal `if` blocks?

Can we collaps those in something like :
```
        if type(args) == list:
            level_error = exp.LevelError(*args)
            _init.assert_called_once_with(ANY, *args)
        else:
            level_error = exp.LevelError(args)
            _init.assert_called_once_with(ANY, args)
```

tests/unit/test_exceptions.py
```diff
@@ -0,0 +1,56 @@
...
+
+from ..unitutil import ANY, initializer_mock
+
+
+class DescribeExceptions(object):
+    @pytest.mark.parametrize("args", ("error", ["hello", "error"], None))
+    def it_can_construct_from_args_level_error(self, request, args):
+        _init = initializer_mock(request, exp.LevelError)
+
+        if type(args) == list:
```
ernestoarbitrio, line 13: In general is not a best practice using `if(s)` in test functions. So i suggest to refactor the test of write different test according to all the cases you have

tests/unit/test_exceptions.py
```diff
@@ -0,0 +1,56 @@
...
+
+        if args:
+            if type(args) == str:
+                assert type(message) == str
+                assert message == args
+            if type(args) == list:
+                assert type(message) == str
+                assert message == list(args)[0]
+        else:
+            assert message is None
```
ernestoarbitrio, line 42: also here

tests/unit/test_exceptions.py
```diff
@@ -0,0 +1,56 @@
...
+            level_error = exp.LevelError(*args)
+        else:
+            level_error = exp.LevelError(args)
+
+        s = str(level_error)
+
+        if not args:
+            assert s == ""
+        else:
+            assert s == level_error.message
```
ernestoarbitrio, line 56: and here

src/histolab/tiler.py
```diff
@@ -28,8 +29,8 @@ class Tiler(Protocol):
     tile_size: int
 
     @lru_cache(maxsize=100)
-    def box_mask(self, slide: Slide) -> sparse._coo.core.COO:
-        """Return binary mask at level 0 of the box to consider for tiles extraction.
+    def box_mask_thumb(self, slide: Slide) -> np.ndarray:
```
ernestoarbitrio, line 32: I'm not sure about the name of this method. It's not meaningful to me ... why did you change `box_mask` onto this new name?

src/histolab/tiler.py
```diff
@@ -485,3 +479,117 @@ def _random_tiles_generator(self, slide: Slide) -> (Tile, CoordinatePair):
...
+
+        for tiles_counter, (score, tile_wsi_coords) in enumerate(highest_score_tiles):
+            tile = slide.extract_tile(tile_wsi_coords, self.level)
+            tile_filename = self._tile_filename(tile_wsi_coords, tiles_counter)
+            tile.save(tile_filename)
+            print(f"\t Tile {tiles_counter} - score: {score} saved: {tile_filename}")
+
+        print(f"{tiles_counter+1} Grid Tiles have been saved.")
+
+    def _scores(self, slide: Slide) -> List[Tuple[float, CoordinatePair]]:
```
ernestoarbitrio, line 542: pleare order property and methods in alphabetical order

src/histolab/tiler.py
```diff
@@ -159,7 +134,7 @@ def level(self) -> int:
     @level.setter
     def level(self, level_: int):
```
ernestoarbitrio, line 135: do not mix private and public methods/properties. In a class public methods should go first then the privates and in each section (priv and pub) they should be ordered alphabetically

src/histolab/tiler.py
alessiamarcolini: Before there were both `box_mask` (mask at level 0) and `box_mask_lvl` (mask at level of extraction) methods. Now there is only the method to return the mask at thumbnail level and I didn't want to keep the same method name and return a completely different thing... but if you think that we could keep only `box_mask` with a different meaning, I'm okay with that

ernestoarbitrio: considering that now we have only 1 name i'd prefer `box_mask` but TBH it's up to you :D

alessiamarcolini: ok changed in #128

nicolebussola (approved):

PR #133, opened 2020-09-03
Grooming and cleaning stuff

ernestoarbitrio:
Cleaning stuff and ordering properties. Just an style exercise PR :D

### DIFF
--- a/src/histolab/slide.py
+++ b/src/histolab/slide.py
@@ -121,7 +121,7 @@ class Slide(object):
             Image containing the selected tile.
         """
 
-        if not self._valid_coordinates(coords):
+        if not self._has_valid_coords(coords):
             # OpenSlide doesn't complain if the coordinates for extraction are wrong,
             # but it returns an odd image.
             raise ValueError(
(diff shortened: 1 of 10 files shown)


src/histolab/slide.py
```diff
@@ -455,6 +431,26 @@ def _thumbnail_size(self) -> Tuple:
             ]
         )
 
+    def _valid_coordinates(self, coords: CoordinatePair) -> bool:
```
alessiamarcolini, line 434: I'm not sure removing the verb helps in conveying the right message: to me, here it seems you are getting new valid coordinates, instead of validating them

alessiamarcolini (commented):
comment on method name change

src/histolab/slide.py
ernestoarbitrio: ok i propose something like `has_valid_coords` because it's a method that lives in a class so it's referring to an object and it's not a generic function applicable to anything

alessiamarcolini: yep because the coordinates must be valid for that particular slide, ok for `has_valid_coords`

ernestoarbitrio: done

alessiamarcolini (approved):

PR #134, opened 2020-09-05
Increase coverage

ernestoarbitrio:
increase test coverage:
 - 100% tile.py

### DIFF
--- a/src/histolab/tile.py
+++ b/src/histolab/tile.py
@@ -175,4 +175,16 @@ class Tile:
         )
 
     def _tissue_mask(self, filters: imf.Compose) -> np.ndarray:
+        """Return the tissue mask of the tile image
+
+        Parameters
+        ----------
+        imf.Compose
+            Filters composition
+
+        Returns
+        -------
+        np.ndarray
+             Tissue mask array
+        """
         return filters(self._image)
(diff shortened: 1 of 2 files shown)


src/histolab/tile.py
```diff
@@ -175,3 +173,18 @@ def _is_almost_white(self) -> bool:
             np.mean(image_gray_arr.ravel()) > 0.9
             and np.std(image_gray_arr.ravel()) < 0.09
         )
+
+    def _tissue_mask(self, filters: imf.Compose) -> np.ndarray:
```
alessiamarcolini, line 177: In my branch I implemented the `apply_filters` (public) method (https://github.com/histolab/histolab/blob/nuclei-scorer/src/histolab/tile.py#L32)

The behavior is actually different though. My `apply_filters` could internally use this method and then put the result back in a `Tile`. What do you think?

ernestoarbitrio: No objection at all! Maybe this PR goes first because it has no new features. So once we merge this you can rebase your branch and change what do you think is better.

alessiamarcolini: alright

alessiamarcolini (approved):
LGTM

PR #143, opened 2020-09-18
Grooming and cleaning pt 2

ernestoarbitrio:


tests/util.py
```diff
@@ -24,3 +24,17 @@ def load_expectation(expectation_file_name, type_=None):  # pragma: no cover
     else:
         raise Exception("Type format not recognized")
     return expectation_data
+
+
+def load_python_expression(expression_file_name):  # pragma: no cover
```
alessiamarcolini, line 29: why are they called 'expressions'?

ernestoarbitrio: Because we can use every kind of python expressions like dictionaries or even quoted strings or whatever could be assimilated as a python expression

alessiamarcolini: alright

ernestoarbitrio: Annnd we use this terminology in all our code bases at work 🤪

alessiamarcolini (approved):

PR #145, opened 2020-09-19
Grooming and cleaning pt3

ernestoarbitrio:
- fix some odd test assertions
- clean names and fixtures
- add missing main comments in module

### DIFF
--- a/src/histolab/tiler.py
+++ b/src/histolab/tiler.py
@@ -337,7 +337,7 @@ class RandomTiler(Tiler):
         check_tissue: bool = True,
         prefix: str = "",
         suffix: str = ".png",
-        max_iter: int = 1e4,
+        max_iter: float = 1e4,
     ):
 
         super().__init__()
--- a/src/histolab/tiler.py
+++ b/src/histolab/tiler.py
@@ -1,3 +1,21 @@
+# encoding: utf-8
+
+# ------------------------------------------------------------------------
+# Copyright 2020 All Histolab Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------
+
 import csv
 import os
 from abc import abstractmethod
--- a/src/histolab/tiler.py
+++ b/src/histolab/tiler.py
@@ -1,12 +1,12 @@
 import csv
 import os
 from abc import abstractmethod
-from typing import List, Tuple
+from typing import List, Tuple, Union
 
 import numpy as np
 
 from .exceptions import LevelError
-from .scorer import Scorer
+from .scorer import RandomScorer, NucleiScorer
 from .slide import Slide
 from .tile import Tile
 from .types import CoordinatePair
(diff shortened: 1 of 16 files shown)


src/histolab/tiler.py
```diff
@@ -337,7 +355,7 @@ def __init__(
         check_tissue: bool = True,
         prefix: str = "",
         suffix: str = ".png",
-        max_iter: int = 1e4,
+        max_iter: float = 1e4,
```
alessiamarcolini, line 358: Why float? How can you have a float number of iterations?

src/histolab/tiler.py
```diff
@@ -519,7 +537,7 @@ class ScoreTiler(GridTiler):
 
     def __init__(
         self,
-        scorer: Scorer,
+        scorer: Union[NucleiScorer, RandomScorer],
```
alessiamarcolini, line 540: Why having the concrete implementations here instead of the abstract one? Isn't the abstract implementations there for this purpose?

tests/integration/test_tile.py
```diff
@@ -21,7 +21,7 @@ class Describe_Tile:
         ),
     )
     def it_knows_if_is_is_almost_white(self, tile_img, expected_result):
-        coords = CoordinatePair(0, 512, 0, 512)
+        coords = CP(0, 512, 0, 512)
```
alessiamarcolini, line 24: Why shortening the name? To me is not clear what a CP is

tests/unit/test_scorer.py
```diff
@@ -22,11 +15,10 @@ def it_can_construct_randomscorer(self, request):
         score = random_scorer(tile)
 
         assert isinstance(random_scorer, scorer.RandomScorer)
-        assert isinstance(random_scorer, scorer.Scorer)
```
alessiamarcolini, line 25: Why are you deleting this? I think that is important for a scorer to be also a Scorer, i.e. implementing the interface

tests/unit/test_scorer.py
```diff
@@ -50,6 +42,5 @@ def it_can_construct_nuclei_scorer(self, request):
             mask_difference_.call_args_list[0][0][1], COMPLEX_MASK
         )
         assert isinstance(nuclei_scorer, scorer.NucleiScorer)
-        assert isinstance(nuclei_scorer, scorer.Scorer)
```
alessiamarcolini, line 53: Same here

src/histolab/tiler.py
ernestoarbitrio: `1e4` in python is float. So we have to decide to change de default value

src/histolab/tiler.py
ernestoarbitrio: Actually Scorer is only used in the scorer module and so it has to be private, and here we cannot import a private class, and BTW this way is more explicit for now

tests/integration/test_tile.py
ernestoarbitrio: `CoordinatePair` is `CP` only in the tests. I'd prefer a short name here to groom the fixtures TBH

tests/unit/test_scorer.py
ernestoarbitrio: RandomScorer is a Scorer by default considering that its signature is `RandomScorer(Scorer)` isn't it?

src/histolab/tiler.py
ernestoarbitrio: maybe int(1e4) ?

alessiamarcolini: as you prefer actually

src/histolab/tiler.py
alessiamarcolini: To me, Scorer shouldn't be private. It has been created to define the interface (it is a protocol) and to be used as an abstract class for the user too (if one wants to implement their own concrete Scorer) and still be compliant with the type hints

some refs:
https://stackoverflow.com/questions/38791739/function-annotation-for-subclasses-of-abstract-class
https://github.com/python/mypy/pull/2853

ernestoarbitrio: ok gotcha

alessiamarcolini (approved):
LGTM

PR #147, opened 2020-09-22
Add rgb_to_lab image filter

alessiamarcolini:


ernestoarbitrio:
Just a general comment: the new image expectations/fixtures are ~640kb each. My concern is that, increasing the test suite will increate the fixtures and the expectations and the entire software will grown a lot in terms of dimension on disk. So maybe we have to think a way to reduce the size of all the files we put on the repo

alessiamarcolini:
@ernestoarbitrio this can become an issue for real. Since it's not related only to this PR, could you please open an Issue so it doesn't get lost?

nicolebussola (approved):
LGTM

PR #153, opened 2020-10-02
Add Watershed segmentation filter

alessiamarcolini:


src/histolab/filters/morphological_filters_functional.py
```diff
@@ -60,3 +63,32 @@ def remove_small_objects(
...
+        Labelled segmentation mask
+
+    References
+    --------
+    .. [1] Watershed segmentation.
+       https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_watershed.html
+    """
+    distance = sc_ndimage.distance_transform_edt(np_mask)
+    local_maxi = sk_feature.peak_local_max(
+        distance, indices=False, footprint=np.ones((6, 6)), labels=np_mask
```
ernestoarbitrio, line 90: Hardcoding params is quite weird to me! Can we use them as params? Just out of curiosity

alessiamarcolini: yes we could but since this filter is inherently complex (it calls "a lot of things" which I don't have real control over), I don't know how much information we want to "expose". I don't even know how to name those params... @nicolebussola any help?

alessiamarcolini:
@nicolebussola I would use some help with the docstrings :)

src/histolab/filters/morphological_filters_functional.py
nicolebussola: I would parametrize the shape of the footprint (now 6) under the parameter: region_shape

src/histolab/filters/morphological_filters_functional.py
```diff
@@ -60,3 +63,32 @@ def remove_small_objects(
...
+def watershed_segmentation(np_mask: np.ndarray) -> np.ndarray:
+    """Segment and label an binary mask with Watershed segmentation [1]_
+
+    The watershed algorithm treats pixels values as a local topography (elevation).
+
+    Parameters
+    ----------
+    np_mask : np.ndarray
+        Input mask
+
```
nicolebussola, line 77: region_shape: int
the local region within which to search for image peaks is  defined as a squared area region_shape x region_shape

ernestoarbitrio (approved):

nicolebussola (approved):
LGTM

PR #164, opened 2020-10-28
Fixes and tweaks

ernestoarbitrio:
some fixes and typos

### DIFF
--- a/tests/unit/test_tiler.py
+++ b/tests/unit/test_tiler.py
@@ -859,7 +859,7 @@ class Describe_ScoreTiler:
         _tile_filename.side_effect = [
             f"tile_{i}_level2_0-10-0-10.png" for i in range(2)
         ]
-        _save_report = method_mock(request, ScoreTiler, "_save_report")
+        _save_report = method_mock(request, ScoreTiler, "_save_report", autospec=False)
         random_scorer = RandomScorer()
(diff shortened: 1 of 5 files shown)


tests/unit/test_tiler.py
```diff
@@ -859,7 +859,7 @@ def it_can_extract_score_tiles_and_save_report(self, request, tmpdir):
         _tile_filename.side_effect = [
             f"tile_{i}_level2_0-10-0-10.png" for i in range(2)
         ]
-        _save_report = method_mock(request, ScoreTiler, "_save_report")
+        _save_report = method_mock(request, ScoreTiler, "_save_report", autospec=False)
```
alessiamarcolini, line 862: What does 'autospec=False' do?

alessiamarcolini (approved):

tests/unit/test_tiler.py
ernestoarbitrio: If you set autospec=True then the mock with be created with a spec from the object being replaced. All attributes of the mock will also have the spec of the corresponding attribute of the object being replaced. Methods and functions being mocked will have their arguments checked and will raise a TypeError if they are called with the wrong signature.

This was the reason why we cannot put the static method decorator on top of actual staticmethod. We used the autospec true instead of false. This way we can use and mock staticmethods

PR #166, opened 2020-11-02
Add GTEx example for tiles extraction and patient-wise split

alessiamarcolini:
TODO:

- [x] write GTEx folder README and modify description of the script
- [x] add AIDP 
- [x] add split result


examples/GTEx/extract_tile_PW_GTEx.py
```diff
@@ -0,0 +1,275 @@
...
+    for sample_id in tqdm(sample_ids):
+        if f"{sample_id}.svs" not in os.listdir(dataset_dir):
+
+            request = requests.get(f"{URL_ROOT}/{sample_id}")
+            with open(
+                os.path.join(dataset_dir, f"{sample_id}.svs"), "wb"
+            ) as output_file:
+                output_file.write(request.content)
+
+            time.sleep(np.random.randint(60, 100))
```
ernestoarbitrio, line 37: this line seems fishy to me? what's the scope of that `time.sleep`?
Furthermore download files with requests where each file can be huge I would suggest to use the `stream` param along with the download by chunks ... something like:

```python
def download_file(url):
    filename = url.split('/')[-1]
    with requests.get(url, stream=True) as r:
        r.raise_for_status()
        with open(filename, 'wb') as f:
            for chunk in r.iter_content(chunk_size=8192): 
                # If you have chunk encoded response uncomment if
                # and set chunk_size parameter to None.
                #if chunk: 
                f.write(chunk)
    return filename
```

alessiamarcolini: The sleep seems necessary because GTEx servers don't seem very performant... when downloading without sleep 70/255 slides were HTML with 'Service not available' 🤷‍♀️

ernestoarbitrio: It seems no API service available i suppose ... If you need a sleep there means that no Api infrastructure is provided and this could be a really mess. IMHO this script won't work for each ID because there's no validator in the endpoint ... so it seems very weak. Do we have time to take another look and try to find alternative service for this kind of image we're searching for?

examples/GTEx/extract_tile_pw_gtex.py
```diff
@@ -0,0 +1,277 @@
...
+    )
+
+    train_df.to_csv(train_csv_path, index=None)
+    test_df.to_csv(test_csv_path, index=None)
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Retrieve a leakage-free dataset of tiles using a collection of WSI"
+        "from the GTEx repository. The WSIs that will be used for tile extraction are "
```
nicolebussola, line 190: remove 190--193 lines

nicolebussola (approved):

examples/GTEx/README.md
```diff
@@ -0,0 +1,59 @@
+# Create a leakage-free dataset of tiles for GTEx
+
+`extract_tile_pw_gtex.py` is a Python script proposed as reference to retrieve a reproducible dataset of tiles using a collection of WSIs from the [GTEx](https://gtexportal.org/home/) public repository. In particular, it can be easily integrated in deep learning pipeline for computational pathology.
```
ernestoarbitrio, line 3: should we use plural here? -> `deep learning pipeline(s)`

ernestoarbitrio (approved):
Just a comment ... otherwise LGTM :)

examples/GTEx/README.md
alessiamarcolini: ok

PR #171, opened 2020-11-06
Add integration test for a broken SVS file

### LINKED ISSUE #5: Check OpenSlide error generation

alessiamarcolini:
fixes #5


src/histolab/slide.py
```diff
@@ -441,10 +441,10 @@ def _wsi(self) -> Union[openslide.OpenSlide, openslide.ImageSlide]:
         """
         try:
             slide = openslide.open_slide(self._path)
-        except openslide.OpenSlideError:
+        except PIL.UnidentifiedImageError:
             raise openslide.OpenSlideError(
                 "Your wsi has something broken inside, a doctor is needed"
```
ernestoarbitrio, line 446: We're catching a PIL exception and raising an OpenSlide one ? I think that's not consistent. It's too much implicit IMHO. I'd prefer to raise the same exception here.

ernestoarbitrio (approved):
LGTM :)

PR #175, opened 2020-11-13
Move Travis CT -> GitHub Actions

ernestoarbitrio:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,138 @@
...
+            python -m pip install --upgrade pip
+            python -m pip install -e .[testing]
+            python -m pip install flake8
+            python -m pip install pooch
+      - name: Lint with flake8
+        run: |
+            # stop the build if there are Python syntax errors or undefined names
+            flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+            # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+            flake8 . --count --exit-zero --max-complexity=10 --max-line-length=88 --statistics
```
alessiamarcolini, line 45: why calling flake8 twice with different parameters? what kind of behavior do you want here? 
moreover, what happens if we change the `.flake8`  conf file?

alessiamarcolini (approved):
LGTM

PR #177, opened 2020-11-16
Make SlideSet iterable

ernestoarbitrio:


src/histolab/slide.py
```diff
@@ -460,6 +460,30 @@ def __init__(
...
+                if os.path.splitext(_path)[1] in self._valid_extensions
+            ]
+        )
+
+    def __len__(self) -> int:
+        """Total number of the slides of this Slideset
+
+        Returns
+        -------
+        int representing the number of the Slides.
```
alessiamarcolini, line 483:
```suggestion
        int 
            number of the Slides.
```

src/histolab/slide.py
```diff
@@ -489,23 +513,9 @@ def save_thumbnails(self, n: int = 0) -> None:
         """
         os.makedirs(self._processed_path, exist_ok=True)
         n = self.total_slides if (n > self.total_slides or n == 0) else n
-        for slide in self.slides[:n]:
+        for slide in list(self.__iter__())[:n]:
             slide.save_thumbnail()
 
-    @lazyproperty
-    def slides(self) -> List[Slide]:
```
alessiamarcolini, line 496: my suggestion was for `SlideSet` to be an iterable object, but I wouldn't remove the `slides` property

alessiamarcolini (commented):
Please give this PR a name, we can link the issue number in the description

src/histolab/slide.py
ernestoarbitrio: why keep that property if slideset contains itself in its __iter__ magic method?

alessiamarcolini: I'm thinking about explicitly getting one slide:
```python
i_th = slideset.slides[i]
```
instead of
```python
i_th = list(slideset)[i]
```

but at this point, we could directly implement `__getitem__`, which I think is the best choice anyway

ernestoarbitrio: yeah exactly i do like it 👍🏽

alessiamarcolini: good, go ahead ;)

src/histolab/slide.py
```diff
@@ -460,6 +460,35 @@ def __init__(
...
+        """
+        return iter(
+            [
+                Slide(os.path.join(self._slides_path, _path), self._processed_path)
+                for _path in os.listdir(self._slides_path)
+                if os.path.splitext(_path)[1] in self._valid_extensions
+            ]
+        )
+
+    def __getitem__(self, slide_id):
```
alessiamarcolini, line 478:
```suggestion
    def __getitem__(self, slide_id: int) -> Slide:
```

alessiamarcolini (approved):
LGTM

PR #179, opened 2020-11-25
Locate tiles on slide's thumbnail

ernestoarbitrio:
This PR contains:
- Annotate slide thumbnail with tiles location. This is a diagnosis method to check where the tiles are (or will be) extracted from the slide.
image

- Remove UserWarning from pytest console and wrap it with pytest context managers

- Various fix


src/histolab/tiler.py
```diff
@@ -67,6 +68,32 @@ def box_mask(self, slide: Slide) -> np.ndarray:
...
+
+        Parameters
+        ----------
+        slide : Slide
+            Slide reference where placing the tiles
+
+        Returns
+        -------
+        img
+            PIL Image with tiles annotations
```
nicolebussola, line 82: PIL Image of the slide thumbnail with the extracted tiles outlined

src/histolab/tiler.py
```diff
@@ -67,6 +68,32 @@ def box_mask(self, slide: Slide) -> np.ndarray:
...
+        """
+        if not os.path.exists(slide.thumbnail_path):
+            slide.save_thumbnail()
+        tiles_coords = (tc[1] for tc in self._tiles_generator(slide))
+        img = Image.open(slide.thumbnail_path)
+        img.putalpha(128)
```
nicolebussola, line 88: Do you think alpha should be a parameter?

src/histolab/tiler.py
```diff
@@ -67,6 +68,32 @@ def box_mask(self, slide: Slide) -> np.ndarray:
...
+        draw = ImageDraw.Draw(img)
+        for coords in tiles_coords:
+            rescaled = np.array(tuple(coords)).reshape(2, 2) / np.power(
+                10, np.ceil(np.log10(slide.dimensions)) - 3
+            ).astype(int)
+            draw.rectangle(tuple(map(tuple, rescaled)), outline="red")
```
nicolebussola, line 94: After some thinking, I agree we should let the user decide the color

nicolebussola (changes requested):
LGTM, just some minor comments

src/histolab/tiler.py
ernestoarbitrio: Considering the wsi(s) color heterogeneity maybe the alpha as param can be really useful !

ernestoarbitrio:
@nicolebussola @alessiamarcolini maybe you both can find a better name than `locate_tiles` ... I don't know maybe ... `tile_outlines` or something else sounds more better in your mind and in this specific topic.

src/histolab/tiler.py
```diff
@@ -67,6 +68,37 @@ def box_mask(self, slide: Slide) -> np.ndarray:
...
+
+        Parameters
+        ----------
+        slide : Slide
+            Slide reference where placing the tiles
+        alpha: int
+            The alpha level to be applied to the slide thumbnail, default to 128.
+        outline: str
+            The outline color for the tile annotations, default to 'red'.
+        Returns
```
alessiamarcolini, line 84:
```suggestion

        Returns
```

src/histolab/tiler.py
```diff
@@ -67,6 +68,37 @@ def box_mask(self, slide: Slide) -> np.ndarray:
...
+        -------
+        img
```
alessiamarcolini, line 86:
```suggestion
        Image
```

src/histolab/tiler.py
```diff
@@ -67,6 +68,37 @@ def box_mask(self, slide: Slide) -> np.ndarray:
...
+        if not os.path.exists(slide.thumbnail_path):
+            slide.save_thumbnail()
+        tiles_coords = (tc[1] for tc in self._tiles_generator(slide))
+        img = Image.open(slide.thumbnail_path)
+        img.putalpha(alpha)
+        draw = ImageDraw.Draw(img)
+        for coords in tiles_coords:
+            rescaled = np.array(tuple(coords)).reshape(2, 2) / np.power(
+                10, np.ceil(np.log10(slide.dimensions)) - 3
+            ).astype(int)
```
alessiamarcolini, line 98: to do this I suggest you use `utils.scale_coordinates()` https://github.com/histolab/histolab/blob/master/src/histolab/util.py#L156

although I may be not getting what you're doing with the `log` and `-3` 🤔

alessiamarcolini (changes requested):
Looks good! 

One thing that I would like to have is the possibility to customize the size of the image returned by `locate_tiles`, let's say we want to include the figure in a scientific paper or in a slide deck, the thumbnail may be too small. So the easiest option would be to accept the scale factor as a parameter and then use the scaled image instead of the thumbnail

src/histolab/tiler.py
ernestoarbitrio: the same we do here
```python 
    @lazyproperty
    def _thumbnail_size(self) -> Tuple:
        return tuple(
            [
                int(s / np.power(10, math.ceil(math.log10(s)) - 3))
                for s in self.dimensions
            ]
        )
```

alessiamarcolini: mmmmmmmm here (lines 96-98) you want to scale the tile coordinates from the dimensions at level 0 to the dimensions of the image you are working on, right? If yes, we can use that function from utils

ernestoarbitrio: ok i'll do that

alessiamarcolini: honestly, I don't remember how we came up with that formula for `_thumbnail_size` 🤨😝

ernestoarbitrio: @alessiamarcolini can you give me some hint on how to use that func? just to understand the params to pass there :D

alessiamarcolini: should be like this
```python
rescaled = scale_coordinates(coords, slide.dimensions, img.size)
```

ernestoarbitrio: thx

ernestoarbitrio:
> Looks good!
> 
> One thing that I would like to have is the possibility to customize the size of the image returned by `locate_tiles`, let's say we want to include the figure in a scientific paper or in a slide deck, the thumbnail may be too small. So the easiest option would be to accept the scale factor as a parameter and then use the scaled image instead of the thumbnail

this will be done in here : #181

alessiamarcolini (approved):
LGTM

src/histolab/tiler.py
```diff
@@ -60,13 +61,48 @@ def box_mask(self, slide: Slide) -> np.ndarray:
...
         raise NotImplementedError
 
+    def locate_tiles(
+        self,
+        slide: Slide,
+        scale_factor: int = 32,
+        alpha: int = 128,
+        outline: str = "red",
+    ) -> PIL.Image.Image:
+        """Place tiles references on the slide thumbnail image
```
alessiamarcolini, line 77: Draw tile box references on a rescaled version of the slide

maybe??

nicolebussola (approved):
LGTM

PR #189, opened 2020-12-04
Augment report using pytest-html and custom helpers

ernestoarbitrio:
Augment test report for TDD actions.

In tests now `expand_tests_report(request, kwargs)` can be used to create a more detailed report like

image

This requires `pytest-html` and the test suite must be ran with this command:

```bash
pytest -svv tests --html=test-report.html --self-contained-html --capture=tee-sys
```

This command will generate an html file with the given name in the directory you're running `pytest` command.

This feature will be very useful for future development when on an integration test we wanna compare the actual and the expected result using real images instead of arrays or expressions.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,3 +9,4 @@ sphinx-rtd-theme
 ipdb
 isort
 sphinxcontrib-katex
+pytest-html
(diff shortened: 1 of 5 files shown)


alessiamarcolini (approved):
LGTM

PR #202, opened 2020-12-30
Raise TileSizeError if the tile size is larger than the slide size

### LINKED ISSUE #192: Tiles extractor stuck when tile size is greater than the slide dimensions
**Bug description**
With a tile extractor like
```
random_tiles_extractor = RandomTiler(
    tile_size=(500, 500),
    n_tiles=6,
    level = 0,
    seed=42,
    check_tissue=False, 
    prefix="random/", 
    suffix=".png")
```
And a `Slide` with dimensions < than 500x500, the `Tiler` get stuck 

**To Reproduce**
1. Instantiate a Slide
2. instantiate a tiler extractor using a tile_size param > than the slide dimensions
3. call `.extract(my_slide)` method

**Expected behavior**
In a case like this throw a meaningful User Error

alessiamarcolini:
fixes #192

AND check in `Tiler.extract()` if the level is available

### DIFF
--- a/src/histolab/tiler.py
+++ b/src/histolab/tiler.py
@@ -25,7 +25,7 @@ from typing import List, Tuple
 import numpy as np
 import PIL
 
-from .exceptions import LevelError
+from .exceptions import LevelError, TileSizeError
 from .scorer import Scorer
 from .slide import Slide
 from .tile import Tile
@@ -188,11 +188,25 @@ class GridTiler(Tiler):
         ----------
         slide : Slide
             Slide from which to extract the tiles
+
+        Raises
+        ------
+        TileSizeError
+            If the tile size is larger than the slide size
         """
         if self.level not in slide.levels:
             raise LevelError(
                 f"Level {self.level} not available. Number of available levels: "
                 f"{len(slide.levels)}"
+            )
+
+        if (
+            self.tile_size[0] > slide.level_dimensions(self.level)[0]
+            or self.tile_size[1] > slide.level_dimensions(self.level)[1]
+        ):
+            raise TileSizeError(
+                f"Tile size {self.tile_size} is larger than slide size "
+                f"{slide.level_dimensions(self.level)} at level {self.level}"
             )
 
         grid_tiles = self._tiles_generator(slide)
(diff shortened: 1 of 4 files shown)


src/histolab/tiler.py
```diff
@@ -182,13 +182,29 @@ def extract(self, slide: Slide):
...
         if self.level not in slide.levels:
             raise LevelError(
                 f"Level {self.level} not available. Number of available levels: "
                 f"{len(slide.levels)}"
             )
 
+        if (
+            self.tile_size[0] > slide.level_dimensions(self.level)[0]
+            or self.tile_size[1] > slide.level_dimensions(self.level)[1]
+        ):
```
ernestoarbitrio, line 202: this expression can be simplified in this way i guess:
```python
if [i for i, j in zip(a, b) if i > j]:
    raise ....
``` 
where `a == self.tile_size` and `b ==slide.level_dimensions(self.level)`

so

```python
if [i for i, j in zip(self.tile_size, slide.level_dimensions(self.level)) if i > j]:
    raise ...
```

ernestoarbitrio (changes requested):
General comment:
considering that you are repeating the same check in each tiler class i'd like to include it in the ` Tiler` class :

```python
def check_something(self):
        return (self.tile_size[0] > slide.level_dimensions(self.level)[0] or self.tile_size[1] > slide.level_dimensions(self.level)[1])
```

and then check for this prop in each subclass

src/histolab/tiler.py
alessiamarcolini: I see where the benefit could be if we had to check more dimensions, but IMHO, since we have only 2 dimensions to check, this simplification is making it harder to understand

ernestoarbitrio: ok

src/histolab/tiler.py
```diff
@@ -135,6 +135,25 @@ def _tile_filename(
...
+        -------
+        bool
+            True if the tile size is smaller or equal than the ``slide`` size at
+            extraction level, False otherwise
+        """
+        return (
+            self.tile_size[0] <= slide.level_dimensions(self.level)[0]
+            and self.tile_size[1] <= slide.level_dimensions(self.level)[1]
+        )
+
```
ernestoarbitrio, line 156: this is not in alphabetical order :D

alessiamarcolini: yes because in the end I changed the method name 🤭

ernestoarbitrio: theeeeen change it 😂

ernestoarbitrio (approved):

Issue #193, opened 2020-12-11
RandomTiler border wackiness extraction


ernestoarbitrio:
**Bug Description**
The RandomTiler extract most of the tiles on the border of the slide. This seems a flaky behaviour and probably depends on the check tissue branches.

PR #203, opened 2020-12-30
Fix has only some tissue using image rather than mask for calculating the variance

ernestoarbitrio:
Fix: #193


tests/unit/test_tile.py
```diff
@@ -239,6 +230,20 @@ def it_knows_how_to_apply_filters_np(self, OtsuThreshold_):
...
+    @pytest.mark.parametrize(
+        "tile_image, expected_value",
+        (
+            (TILES.LIVER_LEVEl2_10907_7808_11707_8608, False),  # all tissue
+            (TILES.LIVER_LEVEl2_20914_13715_21714_14515, True),  # some tissue
+            (TILES.LIVER_LEVEl2_57138_8209_57938_9009, True),  # some tissue
+            (TILES.LIVER_LEVEl2_38626_13514_39426_14315, False),  # no tissue
+        ),
+    )
+    def it_knows_if_it_has_only_some_tissue(self, tile_image, expected_value):
```
alessiamarcolini, line 242: this is an integration test and it should be moved to the correct module

src/histolab/tile.py
```diff
@@ -168,10 +168,9 @@ def _has_only_some_tissue(self, near_zero_var_threshold: float = 0.1) -> np.bool
             True if the image is composed by only some tissue. False if the tile is
             composed by all tissue or by no tissue at all.
         """
-        filters = FiltersComposition(Tile).tissue_mask_filters
-        tissue_mask = filters(self._image)
-
-        return np.var(tissue_mask) > near_zero_var_threshold
+        np_img = np.array(self._image.convert("L"))
+        np_img = 1.0 * (np_img > 150)
+        return np.var(np_img) > near_zero_var_threshold
```
alessiamarcolini, line 173: I see that now the implementation is different with respect to what we discussed, a bit of explanation on why this cutoff at 150?

alessiamarcolini (changes requested):
test to be placed in the correct module

src/histolab/tile.py
```diff
@@ -153,14 +153,14 @@ def tissue_ratio(self) -> float:
...
     # ------- implementation helpers -------
 
-    def _has_only_some_tissue(self, near_zero_var_threshold: float = 150.0) -> np.bool_:
+    def _has_only_some_tissue(self, min_var: float = 150.0) -> np.bool_:
         """Check if the tile is composed by only some tissue.
 
         Parameters
         ----------
-        near_zero_var_threshold : float, optional
+        min_var : float, optional
```
alessiamarcolini, line 161: now the docstring is not correct because it refers to the old filters that were applied

alessiamarcolini (approved):
LGTM

PR #204, opened 2020-12-30
Tilers control tissue_percent parameter of has_enough_tissue

### LINKED ISSUE #163: Tilers should control tissue_percent parameter of has_enough_tissue
At the moment, the user can only decide whether to apply the tissue check during tile extraction or not, and not decide the percentage of tissue required. 

We should add another parameter `tissue_percent` to the tilers' constructor to be used during the `has_enough_tissue` check

alessiamarcolini:
fixes #163

### DIFF
--- a/tests/unit/test_tiler.py
+++ b/tests/unit/test_tiler.py
@@ -619,7 +619,7 @@ class Describe_GridTiler:
             ),
         ),
     )
-    def it_can_generate_random_tiles_with_no_check_tissue(
+    def it_can_generate_grid_tiles_with_no_check_tissue(
         self,
         request,
         tmpdir,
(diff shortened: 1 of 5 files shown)


tests/unit/test_tiler.py
```diff
@@ -455,60 +536,95 @@ def it_can_calculate_n_tiles_row(
...
-                True,
-                [True, False],
-                1,
+                [False, False],
+                2,
             ),
         ),
     )
-    def it_can_generate_grid_tiles(
+    def it_can_generate_random_tiles_with_no_check_tissue(
```
ernestoarbitrio, line 622: i think the name is incorrect, why the test is asking for a grid tiler while the name mentions the random one? Typo?

ernestoarbitrio (changes requested):
justo a typo, then LGTM

ernestoarbitrio (approved):

PR #213, opened 2021-01-10
Implement Filter protocol

### LINKED ISSUE #160: Implement Filters Protocol
It could be nice to implement two Protocols for our Filter classes in the filters subpackage, namely `ImageFilter` and `MorphologicalFilter`.

This would be very similar to the implementation of a Scorer protocol (see https://github.com/histolab/histolab/blob/master/src/histolab/scorer.py#L36-L44)

The idea would be to have an abstract and not implemented `__call__` method and a concrete `__repr__` method, since every filter share the same `__repr__` implementation:
```python
def __repr__(self) -> str:
    return self.__class__.__name__ + "()"
```

Careful attention must be paid in the type annotations for the `__call__` method for the two protocols:
- image filters take a `PIL.Image.Image` and output `Union[PIL.Image.Image, np.ndarray]`
- morphological filters take a `np.ndarray` and output a `np.ndarray`

Once we have the protocols, all the classes must subclass the respective protocol to get the implementation of `__repr__`, and we can use the protocols to improve the type annotations of `apply_filters` method of `Tile` (see https://github.com/histolab/histolab/blob/master/src/histolab/tile.py#L50)

alessiamarcolini:
fixes #160

### DIFF
--- a/src/histolab/filters/image_filters.py
+++ b/src/histolab/filters/image_filters.py
@@ -17,6 +17,7 @@
 # ------------------------------------------------------------------------
 
 import operator
+from abc import abstractmethod
 from typing import Any, Callable, List, Union
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


src/histolab/filters/image_filters.py
```diff
@@ -17,16 +17,45 @@
...
 
 import operator
-from typing import Any, Callable, List, Union
+from abc import abstractmethod
+from typing import Callable, List, Union
 
 import numpy as np
 import PIL
 
 from .. import util as U
```
ernestoarbitrio, line 26: I know that you didn't touch this line, but i wonder why we used a short name like `U` for the util module. Considering that you're touching this file in this PR i'd ask you to change `U` and import and use `utils` as is.

ernestoarbitrio (changes requested):
1 comment ... then LGTM!!!

src/histolab/filters/image_filters.py
alessiamarcolini: done

ernestoarbitrio (approved):

Issue #215, opened 2021-01-18
Refactor ScoreTiler to avoid generating the tiles twice


alessiamarcolini:
Please see PR #212 for context

PR #216, opened 2021-01-18
Tiler refactor

ernestoarbitrio:
This fix: #215 

Furthermore it includes some new refactoring to avoid code fragment duplication and also uniforms names in the Tiler classes

### DIFF
--- a/src/histolab/tiler.py
+++ b/src/histolab/tiler.py
@@ -727,22 +727,6 @@ class ScoreTiler(GridTiler):
         ValueError
             If ``n_tiles`` is negative.
         """  # noqa
-        # scores = (
-        #     self._scale_scores(self._scores(slide)) if scaled else self._scores(slide)
-        # )
-        #
-        # sorted_tiles_by_score = sorted(scores, key=lambda x: x[0], reverse=True)
-        #
-        # if self.n_tiles < 0:
-        #     raise ValueError(f"'n_tiles' cannot be negative ({self.n_tiles})")
-        #
-        # highest_score_tiles = (
-        #     sorted_tiles_by_score[: self.n_tiles]
-        #     if self.n_tiles > 0
-        #     else sorted_tiles_by_score
-        # )
-        #
-        # return highest_score_tiles
         all_scores = self._scores(slide)
         scaled_scores = self._scale_scores(all_scores)
 
--- a/src/histolab/tiler.py
+++ b/src/histolab/tiler.py
@@ -164,8 +164,27 @@ class Tiler(Protocol):
     def _tiles_generator(self, slide: Slide) -> None:
         raise NotImplementedError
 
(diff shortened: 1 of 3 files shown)


src/histolab/tiler.py
```diff
@@ -158,16 +161,35 @@ def _tile_filename(
...
         raise NotImplementedError
 
-    @lazyproperty
-    def _tiles_locator(self):
-        return (
-            self._highest_score_tiles
-            if isinstance(self, ScoreTiler)
-            else self._tiles_generator
-        )
+    def _validate_levels(self, slide: Slide) -> None:
```
alessiamarcolini, line 167: I would prefer to have a method that returns a bool whether the level is valid or not and then let the caller decide what to do with it (it may not want to raise an exception), as we did with `_has_valid_tile_size`. Moreover the method name is not coherent with what it does because we are also checking the tile size here, but with the previous suggestion it should fix itself automatically

src/histolab/tiler.py
```diff
@@ -834,12 +827,12 @@ def _scores(self, slide: Slide) -> List[Tuple[float, CoordinatePair]]:
             List of tuples containing the score and the extraction coordinates for each
             tile. Each tuple represents a tile.
         """
-        if next(self._tiles_generator(slide), None) is None:
+        if next(super()._tiles_generator(slide), None) is None:
```
alessiamarcolini, line 830: why? not needed actually. if you want to be explicit then ok

ernestoarbitrio: nope cause if i don't call the super method it falls into a cricular reference call ... 🗡️

src/histolab/tiler.py
ernestoarbitrio: having a boolean check doesn't reduce the code fragment duplication. Considering that we do the same check into the 3 classes i'd like to have a sort of validator that raises those exception when we call the `extract`

src/histolab/tiler.py
alessiamarcolini: ok I get it

src/histolab/tiler.py
alessiamarcolini: mm ok, since it is internal ok.. anyway we can always refactor it if needed.

anyway `_validate_levels` validates the tile size as well 🤷‍♀️ so I would change the name

ernestoarbitrio: ok ,,, any suggestion on the name?

alessiamarcolini: `_validate_level_tile_size` ?
... we have a method that does two things

ernestoarbitrio: ok maybe we can split it in 2 validators

what do you think?

alessiamarcolini: yes better

alessiamarcolini (approved):
LGTM

PR #226, opened 2021-02-09
Allow pathlib.Path as Slide path paramerer

### LINKED ISSUE #225: Slide accepting pathlib.Path as path of slide on disk
OpenSlide only supports `str` as for path slides, but on our side we could accept `pathlib.Path` as well and convert it to `str` internally.

alessiamarcolini:
fixes #225

### DIFF
--- a/src/histolab/slide.py
+++ b/src/histolab/slide.py
@@ -61,10 +61,7 @@ class Slide:
     def __init__(
         self, path: Union[str, pathlib.Path], processed_path: Union[str, pathlib.Path]
     ) -> None:
-        if isinstance(path, pathlib.Path):
-            path = str(path)
-
-        self._path = path
+        self._path = str(path) if isinstance(path, pathlib.Path) else path
         self._processed_path = processed_path
 
     def __repr__(self):
--- a/src/histolab/slide.py
+++ b/src/histolab/slide.py
@@ -52,13 +52,18 @@ class Slide:
 
     Arguments
     ---------
-    path : str
+    path : Union[str, pathlib.Path]
         Path where the WSI is saved.
-    processed_path : str
+    processed_path : Union[str, pathlib.Path]
         Path where thumbnails and scaled images will be saved to.
     """
 
-    def __init__(self, path: str, processed_path: str) -> None:
+    def __init__(
+        self, path: Union[str, pathlib.Path], processed_path: Union[str, pathlib.Path]
+    ) -> None:
+        if isinstance(path, pathlib.Path):
+            path = str(path)
+
         self._path = path
         self._processed_path = processed_path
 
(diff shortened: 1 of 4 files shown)


tests/unit/test_tiler.py
```diff
@@ -340,12 +341,27 @@ def it_can_generate_random_tiles_even_when_coords_are_not_valid(
...
+            (str),
+            (Path),
+        ],
+    )
+    def it_can_extract_random_tiles(
+        self,
+        request,
+        tmpdir,
+        path_type_transform,
+    ):
```
ernestoarbitrio, line 356: I don't see the need to have both cases here. Considering that this change is a `Slide` change and in here we're testing the tiler so this parametrisation is pointless IMHO. You already have a test with both str and Path in the slides unit tests ... so I'd revert the change here

src/histolab/slide.py
```diff
@@ -52,13 +52,18 @@ class Slide:
...
         Path where thumbnails and scaled images will be saved to.
     """
 
-    def __init__(self, path: str, processed_path: str) -> None:
+    def __init__(
+        self, path: Union[str, pathlib.Path], processed_path: Union[str, pathlib.Path]
+    ) -> None:
+        if isinstance(path, pathlib.Path):
+            path = str(path)
+
```
ernestoarbitrio, line 66: I think this if is not needed. just cast the path to string in any case and we're done 🤷🏽‍♂️

src/histolab/slide.py
```diff
@@ -52,13 +52,18 @@ class Slide:
...
         self._path = path
```
ernestoarbitrio, line 67:
```suggestion
        self._path = str(path)
```

ernestoarbitrio (changes requested):
Some comments left

src/histolab/slide.py
alessiamarcolini: ehm nope
```python
In [1]: str(None)
Out[1]: 'None'
```

tests/unit/test_tiler.py
alessiamarcolini: > You already have a test with both str and Path in the slides unit tests

yes but since we are lazy loading the slide, that test is not meaningful if the goal is checking if histolab can handle `pathlib.Path` before opening a slide with openslide

I'm adding another unit test that opens a slide from disk to check the `._wsi` attribute and I'm reverting this change

src/histolab/slide.py
ernestoarbitrio: ok, seems fair so ...why we don't use a ternary operator in the assigment?
```python
self._path = str(path) if isinstance(path, pathlib.Path) else path
```

ernestoarbitrio (changes requested):
just a further minor comment

src/histolab/slide.py
alessiamarcolini: yeah why not

ernestoarbitrio (approved):
Lgtm

PR #233, opened 2021-02-28
change Callable[..., bool] to Callable[..., Any]

ernestoarbitrio:
here the reason: https://docs.python.org/3/library/operator.html#operator.lt

### DIFF
--- a/src/histolab/util.py
+++ b/src/histolab/util.py
@@ -71,14 +71,14 @@ def np_to_pil(np_img: np.ndarray) -> PIL.Image.Image:
         The image represented as PIL Image
     """
 
-    def _transform_bool(image_array):
-        return image_array.astype(np.uint8) * 255
+    def _transform_bool(img_array: np.ndarray) -> np.ndarray:
+        return img_array.astype(np.uint8) * 255
 
-    def _transform_float(image_array):
+    def _transform_float(img_array: np.ndarray) -> np.ndarray:
         return (
-            image_array.astype(np.uint8)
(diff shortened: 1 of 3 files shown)


src/histolab/util.py
```diff
@@ -206,6 +204,7 @@ def threshold_to_mask(
     np.ndarray
         Boolean NumPy array representing a mask where a pixel has a value True
         if the corresponding input array pixel exceeds the threshold value.
+        if the corresponding input array pixel exceeds the threshold value.
```
alessiamarcolini, line 207: ?

alessiamarcolini (commented):
I'm not really understanding why 😇

ernestoarbitrio:
It's formally wrong . So if in future we wanna integrate a static type checker like mypy this will be marked as an error. Also my ide says that the return type is wrong.

src/histolab/util.py
ernestoarbitrio: there was a blank line, i don't know why github is marking this change this way

ernestoarbitrio:
> I'm not really understanding why this change is needed 😇

left a general comment, furthermore if you read here  https://docs.python.org/3/library/operator.html#operator.lt there is this sentence that makes the things clear:
```
Note that these functions can return any value, which may or may not be interpretable as a Boolean
```

alessiamarcolini:
Oh, now I get it, I probably overlooked that sentence

alessiamarcolini (approved):

PR #236, opened 2021-03-09
Refactor Slide: return thumbnail and scaled image instead of saving them

alessiamarcolini:
Replace `Slide`'s methods `save_thumbnail()` and `save_scaled_image()` wih `thumbnail` and `scaled_image()` in order to return the PIL images instead of saving them to disk.

Moreover, `Tiler`s save tiles in `<processed_path>/<tile_filename>` instead of `<processed_path>/tiles/<tile_filename>`


src/histolab/slide.py
```diff
@@ -79,12 +82,12 @@ def biggest_tissue_box_mask(self) -> np.ndarray:
...
+        mask : np.ndarray
             Binary mask of the box containing the max area of tissue. The dimensions are
             those of the thumbnail.
 
         """
-        thumb = self._wsi.get_thumbnail(self._thumbnail_size)
+        thumb = self.thumbnail
         filters = FiltersComposition(Slide).tissue_mask_filters
 
         thumb_mask = filters(thumb)
```
ernestoarbitrio, line 93:
```suggestion
        filters = FiltersComposition(Slide).tissue_mask_filters

        thumb_mask = filters(self.thumbnail)
```

src/histolab/slide.py
```diff
@@ -205,9 +210,7 @@ def locate_biggest_tissue_box(
             PIL Image of the rescaled slide with the biggest tissue bounding box
             outlined.
         """
-        if not os.path.exists(self.scaled_image_path(scale_factor)):
-            self.save_scaled_image(scale_factor)
-        img = PIL.Image.open(self.scaled_image_path(scale_factor))
+        img = self.scaled_image(scale_factor)
```
ernestoarbitrio, line 213: i don't like to replace `img` some line further, can we just pass `self.scaled_image(scale_factor)` to the filters at line 217?

src/histolab/slide.py
```diff
@@ -476,7 +433,7 @@ def _resampled_dimensions(
 
         Returns
         -------
-        tuple
+        Tuple[int, int, int, int]
```
ernestoarbitrio, line 436: I'll add the x and y references in the docstring like x_upper etc...

src/histolab/slide.py
```diff
@@ -580,35 +539,49 @@ def __len__(self) -> int:
 
     # ---public interface methods and properties---
 
-    def save_scaled_slides(self, scale_factor: int = 32, n: int = 0) -> None:
-        """Save rescaled images
+    def scaled_slides(
+        self, scale_factor: int = 32, n: int = 0
+    ) -> List[PIL.Image.Image]:
+        """Return rescaled slides.
```
ernestoarbitrio, line 545: i don't like this name, actually we're returning a list of pil images not a list of slides, isn't it? Maybe a better name for this would be worthwile

tests/unit/test_slide.py
```diff
@@ -495,8 +358,8 @@ def but_it_raises_error_when_it_doesnt_exist(self):
 
         assert (
             str(err.value)
-            == "Cannot display the slide thumbnail:[Errno 2] No such file or "
-            f"directory: {repr(os.path.join('processed', 'thumbnails', 'b.png'))}"
+            == "Cannot display the slide thumbnail: The wsi path resource doesn't "
+            "exist: a/b"
```
ernestoarbitrio, line 362: what's a/b?

ernestoarbitrio (changes requested):
left some comments, i'll take another look after your answers :D

src/histolab/slide.py
```diff
@@ -266,37 +269,16 @@ def resampled_array(self, scale_factor: int = 32) -> np.array:
...
-        os.makedirs(self._processed_path, exist_ok=True)
-
-        img = self._wsi.get_thumbnail(self._thumbnail_size)
-
-        folder = os.path.dirname(self.thumbnail_path)
-        pathlib.Path(folder).mkdir(exist_ok=True)
-        img.save(self.thumbnail_path)
-
-    def scaled_image_path(self, scale_factor: int = 32) -> str:
-        """Return slide image path.
```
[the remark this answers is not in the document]
alessiamarcolini: do you want to remove all deprecated things here or in a separate PR?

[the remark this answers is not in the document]
ernestoarbitrio: i think here is safe

src/histolab/slide.py
alessiamarcolini: we need `img` defined before because here https://github.com/histolab/histolab/blob/refactor-slide-save-paths/src/histolab/slide.py#L220 you need to draw a line on it

tests/unit/test_slide.py
alessiamarcolini: this is the path of the wsi

src/histolab/slide.py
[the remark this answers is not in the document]
alessiamarcolini: I couldn't remove `_biggest_regions` of `Slide` because it is needed by `locate_biggest_tissue_box()`. The latter method is going to be removed when we have a method in `BinaryMask` that lets you visualize the outline of the extraction mask on the slide.

tests/unit/test_slide.py
ernestoarbitrio: ahahaha what a dork

src/histolab/slide.py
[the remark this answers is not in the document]
ernestoarbitrio: mmm what do you suggest? Wanna do this in another PR or wanna you do this change in here?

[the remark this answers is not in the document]
alessiamarcolini: I would do this in a separate PR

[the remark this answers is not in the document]
ernestoarbitrio: alright ... let's do this in a furhter one

PR #237, opened 2021-03-13
Use logger object instead of logging module when logging tiler updates

kheffah:
This is a simple bug fix. Currently the `tile` module logger uses the logging module instead of the logger object, which results in non-specific logger messages from the **root**:
``INFO:root:	 Tile 0 - score: 0.5016428618605987 saved: vlres_tiles/tile_0_level3_38914-40962-40962-43010.png``

Instead of specific messages that bear the **tiler** prefix, i.e.:
``INFO:tiler:	 Tile 0 - score: 0.5016428618605987 saved: vlres_tiles/tile_0_level3_38914-40962-40962-43010.png``

When this is part of a large codebase, it is desirable to know where the log message is coming from. This fixes this issue.

### DIFF
--- a/src/histolab/tiler.py
+++ b/src/histolab/tiler.py
@@ -275,8 +275,8 @@ class GridTiler(Tiler):
             tile_filename = self._tile_filename(tile_wsi_coords, tiles_counter)
             full_tile_path = os.path.join(slide.processed_path, tile_filename)
             tile.save(full_tile_path)
-            logging.info(f"\t Tile {tiles_counter} saved: {tile_filename}")
-        logging.info(f"{tiles_counter} Grid Tiles have been saved.")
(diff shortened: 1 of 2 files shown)


ernestoarbitrio (changes requested):
Hi @kheffah, thank for your contribution. That's awesome. I'd ask you to fix the test expectation to make the CI passes. It's a very easy fix in the `test_tiler` module :D

kheffah:
> Hi @kheffah, thank for your contribution. That's awesome. I'd ask you to fix the test expectation to make the CI passes. It's a very easy fix in the `test_tiler` module :D

Sure thing :)

ernestoarbitrio:
@kheffah We have an issue on the coverall api on forked repos. So I'm gonna merge this PR anyway. We're working to find a solution to the issue. Thanks again for the contribution :D

PR #256, opened 2021-04-24
Fix random choice of coordinates within true values of a binary mask

### LINKED ISSUE #246: Random tile coordinates
when we compute the random coordinate of a tile in the Random [tiler](https://github.com/histolab/histolab/blob/0f0c18c2b43a5b993674323f6eda2b7b50f53a23/src/histolab/tiler.py#L578), based on binary mask, we are considering random coordinates (x,y). Precisely, we are taking independently x and y as random position where the binary mask is True, however (x,y) could be taken in a position where the mask is False. We should instead fix x or y where the binary mask is True and then take the corresponding y or x.

alessiamarcolini:
fixes #246


src/histolab/util.py
```diff
@@ -117,6 +117,25 @@ def polygon_to_mask_array(dims: tuple, vertices: CoordinatePair) -> np.ndarray:
...
+
+    Returns
+    -------
+    Tuple[int, int]
+        Random pair of indices where the ``binary_mask`` is True.
+    """
+    x = np.random.choice(np.where(binary_mask)[0])
+    y = np.random.choice(np.where(binary_mask[x])[0])
+
+    return x, y
```
ernestoarbitrio, line 136: I don't see any reason of a new util functions that is used basically in one single place. If in future we need this in more than 1 code portion so we can introduce it in the `util` module.

ernestoarbitrio (changes requested):
per comments

src/histolab/util.py
alessiamarcolini: alright, should it be a private static method of `RandomTiler` then?

ernestoarbitrio: for 2 lines and then used in only one part in the randomtiler? i don't think so just copy and paste that 2 lines when you need those :)

alessiamarcolini: mmm not so easily testable in that case

ernestoarbitrio: why? you could check the call args and assert that the y has the correct value you are expecting and that depends on the x.

```
scale_coordinates(
            reference_coords=CoordinatePair(x_ul_lvl, y_ul_lvl, x_br_lvl, y_br_lvl),
            reference_size=binary_mask.shape[::-1],
            target_size=slide.dimensions,
)
```

during the test you can assert that the scale_coordinates has been called with the right x_ul_lvl and y_ul_lvl .. do you?

alessiamarcolini: in the test I want to check that the coordinates I sampled correspond to a True value in the mask, and I won't be checking that with your proposal...

tests/unit/test_tiler.py
```diff
@@ -152,6 +152,9 @@ def it_can_generate_random_coordinates(self, request, tmpdir):
 
         _box_mask_thumb.assert_called_once_with(binary_mask, slide)
         _tile_size.assert_has_calls([call((128, 128))])
+        _random_choice_true_mask2d.assert_called_once_with(
+            NpArrayMock.ONES_500X500_BOOL
+        )
         _scale_coordinates.assert_called_once_with(
             reference_coords=CP(x_ul=0, y_ul=0, x_br=128, y_br=128),
             reference_size=(500, 500),
```
ernestoarbitrio, line 160: i don't understand why this is not enough? 
this
```
_random_choice_true_mask2d.return_value = (0, 0)
```
is basically asserting that x and y are 0 and so this line asserts the same thing i believe

alessiamarcolini: here I assert the fact that starting from (0,0) - which I know because I mocked `random_choice_true_mask2d` - I get to (128,128) via the scaling of the tile dimensions and then the sum.. but I'm not really testing if the coordinates UL are correct

ernestoarbitrio: oh i see they're random that's the point. I don't know maybe the integration tests are enough to cover this .... but if you feel more safe this way, i can buy that. So... just let me know and I'll stamp the PR

alessiamarcolini: I would prefer to keep the unit test as it is, via the integration tests we have we never noticed (visually), probably due to a lucky seed

ernestoarbitrio: aight

ernestoarbitrio (approved):

PR #263, opened 2021-04-29
add images to docs, including filters and masks

nicolebussola:


src/histolab/filters/image_filters.py
```diff
@@ -1259,6 +1283,8 @@ class YenThreshold(ImageFilter):
     gray level (s-1). In this filter, pixels below the computed threshold are
     considered as foreground.
 
+    .. figure:: https://user-images.githubusercontent.com/31658006/116542194-ab065800-a8ec-11eb-9fea-24dd97de8226.pn
```
alessiamarcolini, line 1286:
```suggestion
    .. figure:: https://user-images.githubusercontent.com/31658006/116542194-ab065800-a8ec-11eb-9fea-24dd97de8226.png
```

src/histolab/filters/image_filters.py
```diff
@@ -620,6 +630,8 @@ class RagThreshold(ImageFilter):
     the segments, combine similar regions based on threshold value,
     and then output these resulting region segments.
 
+    .. figure:: https://user-images.githubusercontent.com/31658006/116543569-506dfb80-a8ee-11eb-9d59-413b33de8e73.jpeg
```
alessiamarcolini, line 633: this image needs to be done again, as both the text and the image are misaligned between the two steps

alessiamarcolini (changes requested):
as of now we only have the image for half of the filters and only for the bbox mask. I guess for now is ok but at some point we should have the image for all of them

alessiamarcolini (approved):

PR #268, opened 2021-05-18
Fix polygon_to_mask_array return mask shape

### LINKED ISSUE #267: `polygon_to_mask_array` returns mask with inverted shape
**Describe the bug**
`polygon_to_mask_array` in the `util` module returns mask with inverted shape

**To Reproduce**
```python
from histolab.util import polygon_to_mask_array
from histolab.types import CoordinatePair

vertices = CoordinatePair(1, 0, 2, 0)
mask = polygon_to_mask_array(dims=(5, 6), vertices=vertices)

mask.shape # (6, 5) instead of (5, 6)
```

**Expected behavior**
The shape should be `(5, 6)`.

alessiamarcolini:
Fixes #267


ernestoarbitrio:
General comment, why move a polygon to a rectangle for the mask array operation? @alessiamarcolini can you give me a rationale of this choice?

alessiamarcolini:
I changed the naming because a polygon can have more than 4 vertices and we explicitly require a `CoordinatePair` which corresponds to 4 vertices. Probably a more appropriate name would be "quadrilateral" but to me it's a bit ugly... @ernestoarbitrio what do you think?

ernestoarbitrio:
> I changed the naming because a polygon can have more than 4 vertices and we explicitly require a `CoordinatePair` which corresponds to 4 vertices. Probably a more appropriate name would be "quadrilateral" but to me it's a bit ugly... @ernestoarbitrio what do you think?

__gotcha__ , I would say that your name is absolutely appropriate. So let's keep it

src/histolab/masks.py
```diff
@@ -27,7 +27,7 @@
 from .types import Region
 from .util import (
     lazyproperty,
-    polygon_to_mask_array,
+    rectangle_to_mask_array,
```
ernestoarbitrio, line 30: what do you think if we get rid of the suffix `array` ? `rectangle_to_mask` is clear enough to me even if we don't specify the return type of this function in the name itself.

ernestoarbitrio: Also ... we have a well detailed doctring for the return types. What do you think?

ernestoarbitrio (approved):
just single comment, otherwise LGTM

src/histolab/masks.py
alessiamarcolini: mmm yeah make sense, I will remove it

PR #271, opened 2021-06-01
Run tests on wsi loaded from external url via CI

ernestoarbitrio:


.github/workflows/tests.yml
```diff
@@ -10,7 +10,23 @@ on:
...
 jobs:
 
+  wsi-artifacts:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download WSI artifacts
+        id: load-wsis
+        run: |
+          mkdir -p tests/fixtures/external_wsi
+          wget http://openslide.cs.cmu.edu/download/openslide-testdata/Aperio/CMU-3.svs -P tests/fixtures/external_wsi
```
alessiamarcolini, line 20: for now it's ok like this because we only have 1 slide to download, but maybe for the future we should read these URLs from a file

.github/workflows/tests.yml
```diff
@@ -92,3 +113,12 @@ jobs:
           status: ${{ job.status }}
           steps: ${{ toJson(steps) }}
         if: failure()
+
+  delete-wsi-artifacts:
```
alessiamarcolini, line 117: why do we need this?

ernestoarbitrio: for a matter of fairness in favor of github. :D basically the artifacts have a retention time of 1 day, and i don't wanna have that useless thing that cannot be used outside each workflow, so each roud i'll delete it

.github/workflows/tests.yml
ernestoarbitrio: +1

alessiamarcolini (approved):

PR #275, opened 2021-06-24
Run build on all branches

alessiamarcolini:


ernestoarbitrio (commented):
Can you explain why this change? I don't see any needs for running tests on other branches different from master

alessiamarcolini:
When we were using TravisCI this was the behavior, then when we switched to GH we dropped it. Was this intentional? I noticed this while updating the supplementary material of the histolab paper

ernestoarbitrio:
> When we were using TravisCI this was the behavior, then when we switched to GH we dropped it. Was this intentional? I noticed this while updating the supplementary material of the histolab paper

It's intentional cause we do not create ad-hoc branched where we wanna run the CI, the only way to create branch is via a PR and on the PR we have the rule on the workflow file

alessiamarcolini:
Alright, I will close this PR

PR #276, opened 2021-06-24
Docs fixes about Slide's processed_path

alessiamarcolini:


src/histolab/slide.py
```diff
@@ -210,12 +210,12 @@ def name(self) -> str:
 
     @lazyproperty
     def processed_path(self) -> str:
-        """Retrieve the path to store processed files generated from the slide.
+        """Retrieve the path to store the tiles generated from the slide.
```
ernestoarbitrio, line 213: I think that considering that this is a lazy property the docstring shouldn't contain the word Retrieve cause it doesn't retrieve anything it's a property

ernestoarbitrio (changes requested):
Per comments

src/histolab/slide.py
```diff
@@ -200,7 +200,7 @@ def locate_mask(
 
     @lazyproperty
     def name(self) -> str:
-        """Retrieve the slide name without extension.
+        """Return the slide name without extension.
```
ernestoarbitrio, line 203: A property doesn't compute any action it's a property and for this the docstring should be something without the verbs. I'd use Slide name without extension.

src/histolab/slide.py
```diff
@@ -210,12 +210,12 @@ def name(self) -> str:
 
     @lazyproperty
     def processed_path(self) -> str:
-        """Retrieve the path to store processed files generated from the slide.
+        """Return the path to store the tiles generated from the slide.
```
ernestoarbitrio, line 213: Same

ernestoarbitrio (changes requested):
Last iteration sorry 🙂

alessiamarcolini:
@ernestoarbitrio I updated all the docstrings of lazyproperties, requested review

ernestoarbitrio (approved):
LGTM :)

PR #300, opened 2021-07-23
Add RAG threshold parameters to allow return labels and masking

kheffah:
## Description

Flexible rag filter allowing: 
1. Masked filtering to exclude, for example, areas outside the tissue
2. Returning labeled mask for individual clustered segments instead of a colored RGB. That way, the returned clusters can be used in downstream analysis.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

None

## Checklist


tests/integration/test_image_filters.py
```diff
@@ -612,6 +612,37 @@ def test_rag_threshold_filter_on_rgb_image(pil_image, expected_image):
...
+            RGB.DIAGNOSTIC_SLIDE_THUMB_RGB,
+            "mask-arrays/diagnostic-slide-thumb-rgb-rag-threshold-maskedlabels",
+            True,
+        ),
+    ),
+)
+def test_rag_threshold_filter_return_labels(pil_image, expected_image, masked):
+    expected_value = load_expectation(expected_image, type_="npy")
+
+    if masked:
```
alessiamarcolini, line 633: Conditionals in unit tests are considered a bad practice. See here: https://testsmells.org/pages/testsmells.html

> Test methods need to be simple and execute all statements in the production method. Conditions within the test method will alter the behavior of the test and its expected output, and would lead to situations where the test fails to detect defects in the production method since test statements were not executed as a condition was not met. Furthermore, conditional code within a test method negatively impacts the ease of comprehension by developers.

To avoid using conditionals you can refactor your test using parametrization or by defining multiple test methods.

In this case you could use parametrization, create the mask beforehand and store it as a fixture in `tests/fixtures/mask-arrays` and then use it as we use it here: https://github.com/histolab/histolab/blob/0f0c18c2b43a5b993674323f6eda2b7b50f53a23/tests/integration/test_morphological_filters.py#L14
while for the second case the mask is simply `None`

alessiamarcolini (changes requested):
Some comments about docstring style and testing best practices

tests/integration/test_image_filters.py
kheffah: OK done.

kheffah:
@alessiamarcolini OK, all done 👍🏻

alessiamarcolini:
@kheffah could you please squash all the "update" commit into a single one? Thanks!!

PS there is a way to add all the suggestions to a single commit from the interface directly

kheffah:
@alessiamarcolini Oops I didn't know that. OK, now commit aaf4a098cf3e21ea1317921a22e250ef747b09cb contains the squashed changes.

alessiamarcolini:
Hi @kheffah, sorry I meant to do an interactive rebase to rewrite history and have only the relevant commits (with a proper message) to have a clean history. Please let me know if you need any help with this :)

kheffah:
> Hi @kheffah, sorry I meant to do an interactive rebase to rewrite history and have only the relevant commits (with a proper message) to have a clean history. Please let me know if you need any help with this :)

Yes please, I've never done that before I'm dealing with a deadline at the moment so maybe at least a few days till I get to it. If you prefer to get it done quickly, feel free to do it :)

kheffah:
@alessiamarcolini OK I officially give up :D Everytime I do the interactive rebase locally it get a conflict, asks me to resolve it, asks me to pull, and then when the rebase succeeds and I push, I still see the commits. I could probably fix this if I spend more time on it but I need have high priority research tasks. But feel free to do it if you'd like.

src/histolab/filters/image_filters.py
```diff
@@ -698,13 +698,20 @@ class RagThreshold(ImageFilter):
...
     Returns
     -------
-    PIL.Image.Image
+    PIL.Image.Image, if not ``return_labels``
         Each segment has been colored based on the average
         color for that segment (and similar segments have been combined).
 
+    np.ndarray, if ``return_labels``
+        Value denotes segment membership.
+
```
ernestoarbitrio, line 714: I'm concerned on the multiple return type of this filter. This can be a problem when will be the moment of using any static type checker like `mypy` or similar. Maybe a solution could be split this filter into 2 filters? I don't know it's just a sort of rubberduck thinking 🤷🏽‍♂️ . Let me know what do you think @kheffah @alessiamarcolini

src/histolab/filters/image_filters.py
```diff
@@ -715,14 +722,30 @@ class RagThreshold(ImageFilter):
...
         self.threshold = threshold
+        self.return_labels = return_labels
 
-    def __call__(self, img: PIL.Image.Image) -> PIL.Image.Image:
-        return F.rag_threshold(img, self.n_segments, self.compactness, self.threshold)
+    def __call__(
+        self,
+        img: PIL.Image.Image,
+        mask: np.ndarray = None,
+    ) -> PIL.Image.Image:
```
ernestoarbitrio, line 740: if we agree with the current implementation, this is not `True` anymore. We should have `Union[ndarray, PIL.Image]`

src/histolab/filters/image_filters_functional.py
```diff
@@ -380,6 +380,8 @@ def rag_threshold(
     n_segments: int = 800,
     compactness: float = 10.0,
     threshold: int = 9,
+    mask: np.ndarray = None,
+    return_labels: bool = False,
 ) -> PIL.Image.Image:
```
ernestoarbitrio, line 385: same here

tests/integration/test_image_filters.py
```diff
@@ -612,6 +612,37 @@ def test_rag_threshold_filter_on_rgb_image(pil_image, expected_image):
     )
 
 
+def _create_rag_mask(pil_image):
+    mask = np.ones((pil_image.size[1], pil_image.size[0]))
+    mask[:100, :] = 0
+    mask[:, :100] = 0
+    mask[-100:, :] = 0
+    mask[:, -100:] = 0
+    return mask
```
ernestoarbitrio, line 621: I'd move this in the test function line 640. If we will use it in other places we should move it on top of the file right after the imports

tests/integration/test_image_filters.py
```diff
@@ -612,6 +612,37 @@ def test_rag_threshold_filter_on_rgb_image(pil_image, expected_image):
...
+
+
+@pytest.mark.parametrize(
+    "pil_image, expected_image, mask",
```
ernestoarbitrio, line 625: can you put the expectation at the end?
```
"pil_image, mask, expected_image"
```
Thanks
Just to be compliant with the other tests

tests/integration/test_image_filters.py
```diff
@@ -612,6 +612,37 @@ def test_rag_threshold_filter_on_rgb_image(pil_image, expected_image):
...
+            RGB.DIAGNOSTIC_SLIDE_THUMB_RGB,
+            "mask-arrays/diagnostic-slide-thumb-rgb-rag-threshold-maskedlabels",
+            _create_rag_mask(RGB.DIAGNOSTIC_SLIDE_THUMB_RGB),
+        ),
+    ),
+)
+def test_rag_threshold_filter_return_labels(pil_image, expected_image, mask):
+    expected_value = load_expectation(expected_image, type_="npy")
+
+    ragged = imf.rag_threshold(pil_image, 650, 20.6, 9, return_labels=True, mask=mask)
```
alessiamarcolini, line 642:
```suggestion
    ragged = imf.rag_threshold(pil_image, 650, 20.6, 9, return_labels=True, mask=mask)
    
```

we have a convention: between the fixtures definition, the function/method call and the assertions we have a blank line :)

ernestoarbitrio (changes requested):
per comments

src/histolab/filters/image_filters_functional.py
alessiamarcolini, line 385:
```suggestion
) -> Union[PIL.Image.Image, np.ndarray]:
```

src/histolab/filters/image_filters.py
```diff
@@ -715,14 +722,30 @@ class RagThreshold(ImageFilter):
     """  # noqa
 
     def __init__(
-        self, n_segments: int = 800, compactness: float = 10.0, threshold: int = 9
+        self,
+        n_segments: int = 800,
+        compactness: float = 10.0,
+        threshold: int = 9,
+        return_labels: bool = False,
     ) -> PIL.Image.Image:
```
alessiamarcolini, line 730:
```suggestion
    ) -> None:
```

src/histolab/filters/image_filters.py
alessiamarcolini, line 740:
```suggestion
    ) -> Union[PIL.Image.Image, np.ndarray]:
```

alessiamarcolini (changes requested):
Last stylistic changes :)

PS after my rebase (which rewrote history) is better for you if you work on a different clone of the repo (only required for upcoming work on this branch)

src/histolab/filters/image_filters.py
alessiamarcolini: yeah I put it on my suggestions... we were reviewing at the same time

ernestoarbitrio:
@kheffah I think after these last comments from me and @alessiamarcolini this PR should be ready to go :D Thanks for your work

alessiamarcolini (commented):
I also asked a review to @nicolebussola if she has time ;)

ernestoarbitrio:
> I also asked a review to @nicolebussola if she has time ;)

Oh ... got it, but I think 2 reviews are more than enough 😄

src/histolab/filters/image_filters_functional.py
```diff
@@ -380,7 +380,9 @@ def rag_threshold(
     n_segments: int = 800,
     compactness: float = 10.0,
     threshold: int = 9,
-) -> PIL.Image.Image:
+    mask: np.ndarray = None,
+    return_labels: bool = False,
+) -> Union[PIL.Image.Image, np.ndarray]:
```
ernestoarbitrio, line 385: Union is not imported in the moule. You should import it ... also I would suggest to run test locally before pushing in order to be sure that all the tests are running locally 😺

kheffah:
@alessiamarcolini @ernestoarbitrio Thanks! All suggestions incorporated now.

alessiamarcolini (approved):
LGTM

nicolebussola (approved):
Nice improvement

PR #301, opened 2021-07-25
Support fetching exact MPP resolutions

kheffah:
This PR supports fetching exact micron-per-pixel (MPP) resolutions from the WSI using [large_image](https://github.com/girder/large_image) as an optional dependency. The rationale here is that different WSIs have different encodings for the internal levels and different scan resolutions. `large_image` handles this by parsing the internal metadata for a wide range of formats beyond those supported by openslide and interpolating between levels when needed to fetch exact MPP resolutions independent of the slide format, scan resolution, manufacturer etc. 

If the user doesn't have `large_image` installed but asks for an mpp resolution, they are prompted to install it. 

@alessiamarcolini @ernestoarbitrio @nicolebussola This is the key feature from the forked version. I added tests for (hopefully) everything, **but they are skipped if `large_image` is not installed**. So if you'd like all the tests to run to maintain coverage, I suggest adding the `large_image` installation to test environment.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

## Checklist


src/histolab/slide.py
```diff
@@ -73,29 +117,43 @@ def dimensions(self) -> Tuple[int, int]:
...
+            Micron per pixel resolution. Takes precedence over level.
 
         Returns
         -------
         tile : Tile
             Image containing the selected tile.
         """
+        assert (level is not None) or (
+            mpp is not None
+        ), "either level or mpp must be provided!"
```
ernestoarbitrio, line 152: I would change it with a raise exception but also use the inverted logic:
```
If level is None and mpp is none:
    raise ....
```
I think using the `and` is more explicit

src/histolab/slide.py
```diff
@@ -352,13 +454,28 @@ def _resample(self, scale_factor: int = 32) -> Tuple[PIL.Image.Image, np.array]:
         """
 
         _, _, new_w, new_h = self._resampled_dimensions(scale_factor)
-        level = self._wsi.get_best_level_for_downsample(scale_factor)
-        whole_slide_image = self._wsi.read_region(
-            (0, 0), level, self._wsi.level_dimensions[level]
-        )
+        if self._use_largeimage:
+            mg = "magnification"
+            kws = (
```
ernestoarbitrio, line 459: For clarity and to be explicit I would use meaningful variable name, cause `mg` and `kws` can be misinterpreted in future

tests/integration/test_slide.py
```diff
@@ -55,9 +95,23 @@ def it_raises_openslideerror_with_broken_wsi(self):
             slide._wsi
 
         assert isinstance(err.value, PIL.UnidentifiedImageError)
-        assert (
-            str(err.value) == "Your wsi has something broken inside, a doctor is needed"
+        broken_err = "Your wsi has something broken inside, a doctor is needed. "
+        broken_err += LARGEIMAGE_INSTALL_PROMPT
```
ernestoarbitrio, line 99: collapse these 2 lines into a single 1 using fstrings for example

tests/integration/test_slide.py
```diff
@@ -55,9 +95,23 @@ def it_raises_openslideerror_with_broken_wsi(self):
...
+
+        with pytest.raises(HistolabException) as err:
+            slide._path = None
+            slide._wsi
+
+        assert isinstance(err.value, HistolabException)
+        broken_err = (
+            "ArgumentError(\"argument 1: <class 'TypeError'>: Incorrect type\")"
         )
+        broken_err += f"\n{LARGEIMAGE_INSTALL_PROMPT}"
```
ernestoarbitrio, line 113: I still don't understand the need of this constant sorry

ernestoarbitrio (changes requested):
Ok I ran out of time. I'll take another look after these comments will be addressed 😄

src/histolab/slide.py
```diff
@@ -26,14 +26,30 @@
...
 from .tile import Tile
 from .types import CoordinatePair
 from .util import lazyproperty
 
+try:
+    import large_image
+    from io import BytesIO
+
+    LARGEIMAGE_IS_INSTALLED = True
+    LARGEIMAGE_INSTALL_PROMPT = ""
```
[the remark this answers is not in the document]
kheffah: Sure thing. These two constants have values that change depending on whether or not ``large_image`` is installed. This is necessary because the try-except import block is at the top of the file (as it should) and depending on whether or not ``large_image`` is available the error message should be different (see my response to your other comment below).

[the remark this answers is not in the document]
kheffah: OK I removed it and used more specific error messages instead.

kheffah:
> Ok I ran out of time. I'll take another look once these comments are addressed

@ernestoarbitrio Thanks, all addressed.

src/histolab/slide.py
```diff
@@ -431,12 +583,17 @@ def _wsi(self) -> Union[openslide.OpenSlide, openslide.ImageSlide]:
...
+                "Your wsi has something broken inside, a doctor is needed. "
+                + self._usage_requires_largeimage
             )
         except FileNotFoundError:
             raise FileNotFoundError(
                 f"The wsi path resource doesn't exist: {self._path}"
             )
+        except Exception as other_error:
+            msg = other_error.__repr__()
+            msg += f"\n{self._usage_requires_largeimage}"
```
ernestoarbitrio, line 595: same question here

src/histolab/tiler.py
```diff
@@ -113,6 +114,18 @@ def locate_tiles(
...
+        """
+        Set tile size either to requested level or if MPP is requested,
+        set tile size relative to base mpp of slide instead.
+        """
+        if self.mpp is None:
+            return
+        sf = self.mpp / slide.base_mpp
+        self.tile_size = tuple(int(j * sf) for j in self.tile_size)
+        if hasattr(self, "pixel_overlap"):
+            self.pixel_overlap = int(self.pixel_overlap * sf)
```
ernestoarbitrio, line 127: I have some concerns on this method:
1. You are setting a Protocol variable in here that basically has some security vulnerabilities because we change the object info
2. You are changing also a shadow name here `pixel_overlap` using the trick of the hasattr... that smells to me. Pixel_overlap doesn't exists in the scope of the `Tiler` object and so it should be set in each Tiler subclass.

I would suggest to remove this method frome here a use setter and getter in each class you need this. If this is not clear I'll try to explain it better. Just let me know

src/histolab/tiler.py
```diff
@@ -276,6 +295,7 @@ def extract(
         level = logging.getLevelName(log_level)
         logger.setLevel(level)
         self._validate_level(slide)
+        self._fix_tile_size_if_mpp(slide)
```
ernestoarbitrio, line 298: see comment above

ernestoarbitrio (changes requested):
Second iteration: some minor comments + 1 major concern. Please let me know if my comments are clear and if you have enough info to go through those. Thanks

src/histolab/slide.py
kheffah: ok, handled through a better error message.

src/histolab/slide.py
```diff
@@ -431,12 +583,17 @@ def _wsi(self) -> Union[openslide.OpenSlide, openslide.ImageSlide]:
             slide = openslide.open_slide(self._path)
         except PIL.UnidentifiedImageError:
             raise PIL.UnidentifiedImageError(
-                "Your wsi has something broken inside, a doctor is needed"
+                "Your wsi has something broken inside, a doctor is needed. "
+                + self._usage_requires_largeimage
```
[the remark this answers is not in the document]
kheffah: OK, handled through a better error message.

src/histolab/slide.py
```diff
@@ -411,13 +545,31 @@ def _thumbnail_size(self) -> Tuple[int, int]:
         Tuple[int, int]
             Thumbnail size
         """
+        if self._use_largeimage:
+            raise ValueError("Please use thumbnail.size instead")
```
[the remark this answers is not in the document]
kheffah: OK, fixed.

src/histolab/tiler.py
kheffah: OK, now the ``Tiler`` protocol only has a getter and does not use hasattr. Setters are used for each of ``GridTiler`` and ``RandomTiler`` to update ``tile_size`` and ``pixel_overlap`` as needed. Test coverage no also ensures ``pixel_overlap`` is successfully updated if ``mpp`` is set.

src/histolab/exceptions.py
```diff
@@ -33,13 +33,17 @@ def __str__(self):
...
+    """Raised when a filter composition for the class is not available"""
+
+
 class LevelError(HistolabException):
     """Raised when a requested level is not available"""
 
 
-class FilterCompositionError(HistolabException):
-    """Raised when a filter composition for the class is not available"""
+class MayNeedLargeImageError(HistolabException):
```
ernestoarbitrio, line 44: Why not only `LargeImageError`?

src/histolab/slide.py
```diff
@@ -64,6 +90,30 @@ def __repr__(self):
 
     # ---public interface methods and properties---
 
+    @lazyproperty
+    def base_mpp(self) -> float:
+        """Get microns-per-pixel resolution at scan magnification."""
```
ernestoarbitrio, line 95: Considering that this is a public property and the docstring becaome automatically part of the doc in read the docs i would add the return statement like
```
Returns
        -------
        bla bla
        """
```

src/histolab/slide.py
```diff
@@ -352,13 +475,28 @@ def _resample(self, scale_factor: int = 32) -> Tuple[PIL.Image.Image, np.array]:
         """
 
         _, _, new_w, new_h = self._resampled_dimensions(scale_factor)
-        level = self._wsi.get_best_level_for_downsample(scale_factor)
-        whole_slide_image = self._wsi.read_region(
-            (0, 0), level, self._wsi.level_dimensions[level]
-        )
+        if self._use_largeimage:
+            magnif = "magnification"
```
ernestoarbitrio, line 479: I think this is not needed, just use `"magnification"` in the kwargs expression

src/histolab/tiler.py
```diff
@@ -276,6 +305,7 @@ def extract(
         level = logging.getLevelName(log_level)
         logger.setLevel(level)
         self._validate_level(slide)
+        self._set_proper_tile_size_and_overlap(slide)
```
ernestoarbitrio, line 308: I think that this line should be:
```
if self.mpp is None:
    tile_size = self.tile_size
    sf = 1.0
else:
    sf = self.mpp / slide.base_mpp
    tile_size = tuple(int(j * sf) for j in self.tile_size)
self.tile_size =  tile_size
self.pixel_overlap = int(sf * self.pixel_overlap)
```

and then you will have 2 setters like:

```
@pixel_overalp.setter
def pixel_overlap(self, value):
    self.pixel_overlap = value

@tile_size.setter
def tile_size(self, value):
   self._tile_size = value
```

This way you could get rid of `_set_proper_tile_size_and_overlap` and `_get_proper_tile_size` and respect one of the python zen rule `Explicit is better than implicit` and `Beautiful is better than ugly` 😉

ernestoarbitrio (changes requested):
Ok I think this should be the last iteration :D

src/histolab/exceptions.py
kheffah: I prefer to keep it explicit as ``LargeImageError`` gives the impression this is coming from the ``large_image`` package whereas in reality it's due to the potential need for large_image to be used.

src/histolab/slide.py
kheffah: ok done.

src/histolab/slide.py
kheffah: ok done.

src/histolab/tiler.py
kheffah: @ernestoarbitrio Actually I prefer to keep it the way it is now to avoid repeating code too much to reduce chances of bugs. The ``_get_proper_tile_size()`` method is in the parent ``Tiler`` module and used by both ``GridTiler`` and ``ScoreTiler`` .. if I get rid of it I'll have to repeat code which makes it harder to maintain and more bug-prone. Also, there's already an existing ``@tile_size.setter``, so this would conflict with adding a new tile size setter. Note that the method ``_set_proper_tile_size_and_overlap`` method is used by both ``GridTiler`` and inherited and used by ``ScoreTiler`` so writing it would will also result in code repetition which could be bug-prone. Another reason to keep ``_set_proper_tile_size_and_overlap`` explicitly is that it can be unit tested as an independent functionality so if something is wrong with the any ``Tiler`` when ``mpp is not None``, we can know if the problem is caused by scale factor differences or some other problem in the ``extract`` method.

ernestoarbitrio: Ok I can buy this for now, but we should think at a better implementation in the future

src/histolab/slide.py
```diff
@@ -48,13 +64,23 @@ class Slide:
     """
 
     def __init__(
-        self, path: Union[str, pathlib.Path], processed_path: Union[str, pathlib.Path]
+        self,
+        path: Union[str, pathlib.Path],
+        processed_path: Union[str, pathlib.Path],
+        use_largeimage=False,
```
ernestoarbitrio, line 70: This argument is missed in the docstring

src/histolab/slide.py
```diff
@@ -64,6 +90,37 @@ def __repr__(self):
...
+        """Get microns-per-pixel resolution at scan magnification.
+
+        Returns
+        -------
+        float
+            Microns-per-pixel resolution at scan (base) magnification.
+
+        """
+        if self._use_largeimage:
+            return self._metadata["mm_x"] * (10 ** 3)
```
ernestoarbitrio, line 104: I think we have an edge case here but maybe I'm wrong. If you add this case `(SVS.TCGA_CR_7395_01A_01_TS1, (128, 128), 0, 2, 10, 0.5)` in `test_extract_tiles_respecting_the_given_tile_size` it will throw a `ValueError` because `self._metadata["mm_x"]` is None ... here the proof:
```
{'levels': 6, 'sizeX': 4443, 'sizeY': 3567, 'tileWidth': 256, 'tileHeight': 256, 'magnification': None, 'mm_x': None, 'mm_y': None}
```

I think we should handle this case. What do you think?

src/histolab/slide.py
kheffah: added.

src/histolab/slide.py
kheffah: ok, handled.

tests/integration/test_slide.py
```diff
@@ -24,6 +25,36 @@ def it_knows_its_name(self):
...
+            ),
+        ],
+    )
+    def it_knows_its_base_mpp(self, use_largeimage, slide_props):
+        slide = Slide(
+            SVS.CMU_1_SMALL_REGION,
+            os.path.join(SVS.CMU_1_SMALL_REGION, "processed"),
+            use_largeimage=use_largeimage,
+        )
+        if slide_props:
```
alessiamarcolini, line 49: instead of having an if here, is it possible to parametrize the slide properties?

tests/integration/test_tiler.py
```diff
@@ -111,35 +111,39 @@ def it_locates_tiles_on_the_slide(
...
+            (TIFF.KIDNEY_48_5, (10, 20), 0, 2, 10, None),
+            (TIFF.KIDNEY_48_5, (20, 10), 0, 20, 20, None),
+            (TIFF.KIDNEY_48_5, (10, 15), 0, 20, 10, None),
         ),
     )
     def test_extract_tiles_respecting_the_given_tile_size(
-        self, tmpdir, fixture_slide, tile_size, level, seed, n_tiles
+        self, tmpdir, fixture_slide, tile_size, level, seed, n_tiles, mpp
     ):
+        use_largeimage = mpp is not None
```
alessiamarcolini, line 137: can you also parametrize this please?

tests/integration/test_tiler.py
```diff
@@ -148,6 +152,83 @@ def test_extract_tiles_respecting_the_given_tile_size(
...
+            tile_size=(128, 128),
+            n_tiles=3,
+            seed=0,
+            check_tissue=True,
+            mpp=0.5,
+        )
+        binary_mask = BiggestTissueBoxMask()
+
+        with pytest.raises(RuntimeError) as err:
+            random_tiles_extractor.final_tile_size = tile_size
```
alessiamarcolini, line 217: in the context manager's body there should be only the code that raises that exception.. so this should go before

tests/unit/test_tiler.py
```diff
@@ -29,18 +30,41 @@
...
+        assert tiler.level == expected_level
+        assert tiler.mpp == expected_mpp
+
+    @pytest.mark.parametrize(
+        "mpp, fixed_tile_size", ((None, (512, 512)), (0.5, (1024, 1024)))
+    )
+    def it_can_fix_tile_size_if_mpp(self, mpp, fixed_tile_size):
+        fake_slide = namedtuple("fake_slide", ["base_mpp"])
+        tiler = RandomTiler((512, 512), 0, True, 80, 0, "", ".png", mpp=mpp)
+        tiler._set_proper_tile_size(fake_slide(0.25))
```
alessiamarcolini, line 64: It took me 5 minutes to understand what was going on here.. 😅 You're faking the value of a property of `Slide` creating the `namedtuple`. IMHO it would be better to use a Slide mock 

However, and more importantly, this is a unit test that is calling `_set_proper_tile_size` and not mocking the internal method that gets called (`_get_proper_tile_size`). Therefore, it's not really testing in isolation. My suggestion would be to have:
1. a unit test that calls and tests the return value of `_get_proper_tile_size`
2. and separately a test for `_set_proper_tile_size` that is basically mocking `_get_proper_tile_size` and testing that the mock is called with the right parameters

Let me know if you have questions or comments 🙏

tests/unit/test_tiler.py
```diff
@@ -408,16 +432,43 @@ def _random_tile_coordinates(self, request):
...
+    @pytest.mark.parametrize(
+        "mpp, fixed_tile_size, fixed_overlap",
+        ((None, (512, 512), 32), (0.5, (1024, 1024), 64)),
+    )
+    def it_can_fix_tile_size_and_overlap_if_mpp(
+        self, mpp, fixed_tile_size, fixed_overlap
+    ):
+        fake_slide = namedtuple("fake_slide", ["base_mpp"])
+        tiler = GridTiler((512, 512), pixel_overlap=32, mpp=mpp)
+        tiler._set_proper_tile_size_and_overlap(fake_slide(0.25))
```
alessiamarcolini, line 467: here the same comment about not testing in isolation

alessiamarcolini (changes requested):

src/histolab/slide.py
```diff
@@ -411,13 +605,37 @@ def _thumbnail_size(self) -> Tuple[int, int]:
         Tuple[int, int]
             Thumbnail size
         """
+        if self._use_largeimage:
+            raise MayNeedLargeImageError(
+                "When use_largeimage is set to True, the thumbnail is fetched "
+                "by the large_image module. Please use thumbnail.size instead."
```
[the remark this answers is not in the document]
ernestoarbitrio: yes this will never happen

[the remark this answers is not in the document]
kheffah: @alessiamarcolini @ernestoarbitrio Sorry for the late response. Actually this does happen and it happened for a couple of cases I worked with. The way `large_image` fetches the thumbnail is different depending on the tile source. If it's a slide compatible with openslide then yes it will most likely be the same, but this will not always be the case. Hence deferring to `large_image` for calculating the size for us.

tests/unit/test_slide.py
```diff
@@ -215,15 +216,11 @@ def it_knows_its_thumbnail_size(self, tmpdir):
 
     def it_raises_error_if_thumbnail_size_and_use_largeimage(self):
         slide = Slide("/a/b/foo", "processed", use_largeimage=True)
-
-        with pytest.raises(MayNeedLargeImageError) as err:
+        with pytest.raises(TileSourceException) as err:
```
alessiamarcolini, line 267: missing blank line

tests/unit/test_tiler.py
```diff
@@ -36,11 +36,11 @@ def it_constructs_from_args(self, level, mpp, request):
         _init = initializer_mock(request, RandomTiler)
 
         random_tiler = RandomTiler(
-            (512, 512), 10, level, 7, True, "", ".png", int(1e4), mpp=mpp
+            tile_size=(512, 512), n_tiles=10, level=level, mpp=mpp
```
alessiamarcolini, line 38: why are we removing parameters?

tests/unit/test_slide.py
```diff
@@ -104,28 +105,51 @@ def it_knows_its_name(self, slide_path, expected_value):
...
+    def but_it_raises_large_image_error_with_unknown_mpp_without_largeimage(
+        self, request
+    ):
+        slide = Slide("foo", "bar", use_largeimage=False)
+        property_mock(request, Slide, "properties", return_value={})
 
         with pytest.raises(MayNeedLargeImageError) as err:
             slide.base_mpp
 
-        assert isinstance(err.value, MayNeedLargeImageError)
```
alessiamarcolini: why removing this?

tests/unit/test_slide.py
```diff
@@ -104,28 +105,51 @@ def it_knows_its_name(self, slide_path, expected_value):
...
-            tmpdir, PILIMG.RGBA_COLOR_500X500_155_249_240, use_largeimage=True
-        )
+    def and_it_raises_value_error_with_unknown_mpp_with_largeimage(self, request):
+        slide = Slide("foo", "bar", use_largeimage=True)
+        property_mock(request, Slide, "_metadata", return_value={})
 
         with pytest.raises(ValueError) as err:
             slide.base_mpp
 
-        assert isinstance(err.value, ValueError)
```
alessiamarcolini: why removing this?

tests/unit/test_slide.py
```diff
@@ -159,15 +183,39 @@ def it_knows_its_dimensions(self, tmpdir):
...
+            "you requested the right `mpp` and/or `tile_size`."
         )
 
+    def it_raises_error_if_bad_args_for_extract_tile(self):
+        slide = Slide("foo", "bar", use_largeimage=True)
+
         with pytest.raises(ValueError) as err:
             slide.extract_tile(CP(0, 10, 0, 10), (10, 10), level=None, mpp=None)
 
-        assert isinstance(err.value, ValueError)
```
alessiamarcolini: same here

tests/unit/test_slide.py
```diff
@@ -159,15 +183,39 @@ def it_knows_its_dimensions(self, tmpdir):
 
         assert slide_dims == (500, 500)
 
-    def it_raises_error_if_bad_args_for_extract_tile(self, tmpdir):
-        slide, _ = base_test_slide(
-            tmpdir, PILIMG.RGBA_COLOR_500X500_155_249_240, use_largeimage=True
+    def it_extracts_tile_with_mpp_without_image_resizing(self, tmpdir, request):
+        slide = Slide(SVS.CMU_1_SMALL_REGION, tmpdir, use_largeimage=True)
+        property_mock(request, PIL.Image.Image, "size", return_value=(128, 128))
```
alessiamarcolini, line 188: shouldn't we assert that the mocks are called? (comment is referring also to other tests)

tests/unit/test_tiler.py
ernestoarbitrio: cause they're the default one and not mandatory, so for sake of readability i'd remove them and use kwargs instead of positional

tests/unit/test_slide.py
ernestoarbitrio: if we use a context manager that check for an exception it's already tested that we are raising that kind of exception. This assert is unnecessary

tests/unit/test_slide.py
ernestoarbitrio: same

tests/unit/test_slide.py
ernestoarbitrio: same

tests/unit/test_slide.py
ernestoarbitrio: not for properties cause we are actually using a real slide

alessiamarcolini: mmm, so what?

ernestoarbitrio: honestly this `extract_tile` is becoming huge. The more i see it the more I'd like to refactor it. So for now i would keep this test as is, it's kinda of weird but it's the only thing I can imagine today. Let's discuss further in the next future. What do u think?

src/histolab/slide.py
```diff
@@ -76,42 +167,103 @@ def dimensions(self) -> Tuple[int, int]:
...
                 f"{self.dimensions}"
             )
 
-        image = self._wsi.read_region(
-            location=(coords.x_ul, coords.y_ul), level=level, size=tile_size
-        )
-        tile = Tile(image, coords, level)
-        return tile
+        if mpp is None:
+            image = self._wsi.read_region(
```
alessiamarcolini, line 224: This will output RGBA images, while if we extract with large_image it will be converted to RGB (by hand, line 241), we should uniform this... I personally would keep it as it was originally, so RGBA

tests/unit/test_tiler.py
```diff
@@ -409,16 +424,29 @@ def _random_tile_coordinates(self, request):
...
+            ANY, (512, 512), level, True, 80, 0, "", ".png", mpp=mpp
+        )
         assert isinstance(grid_tiler, GridTiler)
         assert isinstance(grid_tiler, Tiler)
 
+    @pytest.mark.parametrize(
+        "level, mpp, expected_level, expected_mpp",
+        ((2, None, 2, None), (None, 0.5, 0, 0.5), (2, 0.5, 0, 0.5)),
+    )
+    def mpp_supercedes_level(self, level, mpp, expected_level, expected_mpp):
```
alessiamarcolini, line 443: with this name this test will not be collected

alessiamarcolini (changes requested):
very last review round ;p

tests/unit/test_slide.py
alessiamarcolini: yes agreed, it's huge but let's merge as it is

tests/unit/test_tiler.py
ernestoarbitrio: i changed it ... it wasn't me :D

src/histolab/slide.py
ernestoarbitrio: are you suggesting `image = image.convert("RGBA")` ?

tests/unit/test_tiler.py
ernestoarbitrio: actually we had 2 i changed only 1 XD

src/histolab/slide.py
alessiamarcolini: I'm suggesting to remove line 241, if that works

tests/unit/test_tiler.py
alessiamarcolini: lol

alessiamarcolini (approved):

ernestoarbitrio (approved):

PR #304, opened 2021-07-26
Finer control of locate_tiles

kheffah:
Finer control of locate_tiles includin

1. Allowing user to pass tiless to avoid re-extraction
2. Optionally not fading image
3. Passing individual colors to specific tiles (eg to emphasize specific regions of interest).

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/src/histolab/tiler.py
+++ b/src/histolab/tiler.py
@@ -20,7 +20,7 @@ import csv
 import logging
 import os
 from abc import abstractmethod
-from typing import List, Tuple
+from typing import Iterable, List, Tuple, Union
 
 import numpy as np
 import PIL
@@ -73,7 +73,9 @@ class Tiler(Protocol):
         extraction_mask: BinaryMask = BiggestTissueBoxMask(),
         scale_factor: int = 32,
         alpha: int = 128,
-        outline: str = "red",
+        outline: Union[str, List] = "red",
+        linewidth: int = 1,
+        tiles: Iterable = None,
     ) -> PIL.Image.Image:
         """Draw tile box references on a rescaled version of the slide
 
@@ -90,6 +92,10 @@ class Tiler(Protocol):
             The alpha level to be applied to the rescaled slide, default to 128.
         outline: str
             The outline color for the tile annotations, default to 'red'.
+        linewidth: int
+            Thickness of line used to draw tiles
+        tiles: Iterable
+            Tiles to visualize. Optional, will be extracted if None.
 
         Returns
         -------
@@ -97,18 +103,28 @@ class Tiler(Protocol):
             PIL Image of the rescaled slide with the extracted tiles outlined
         """
         img = slide.scaled_image(scale_factor)
-        img.putalpha(alpha)
+        if alpha is not None:
+            img.putalpha(alpha)
         draw = PIL.ImageDraw.Draw(img)
 
-        tiles = (
-            self._tiles_generator(slide, extraction_mask)[0]
-            if isinstance(self, ScoreTiler)
-            else self._tiles_generator(slide, extraction_mask)
-        )
-        tiles_coords = (tile[1] for tile in tiles)
-        for coords in tiles_coords:
+        if tiles is None:
+            tiles = (
+                self._tiles_generator(slide, extraction_mask)[0]
+                if isinstance(self, ScoreTiler)
+                else self._tiles_generator(slide, extraction_mask)
+            )
+        tiles_coords = [tile[1] for tile in tiles]
+
+        # maybe user specified colors for each tile
+        if isinstance(outline, str):
+            outline = [outline] * len(tiles_coords)
+
+        for i, coords in enumerate(tiles_coords):
             rescaled = scale_coordinates(coords, slide.dimensions, img.size)
-            draw.rectangle(tuple(rescaled), outline=outline)
+            clr = outline[i]
+            if not isinstance(clr, str):
+                clr = tuple([int(255 * j) for j in clr])
+            draw.rectangle(tuple(rescaled), outline=clr, width=linewidth)
         return img
 
     # ------- implementation helpers -------
--- a/tests/unit/test_tiler.py
+++ b/tests/unit/test_tiler.py
@@ -4,6 +4,7 @@ import os
 import re
 from unittest.mock import call
 
+import numpy as np
 import pytest
 
 from histolab.exceptions import LevelError, TileSizeError
@@ -776,6 +777,53 @@ class Describe_GridTiler:
 
         assert type(coords_within_extraction_mask) == bool
         assert coords_within_extraction_mask == expected_result
+
+    @pytest.mark.parametrize(
+        "alpha, pass_tiles, outline, expected_topleft",
+        (
+            (128, False, "red", (255, 255, 155, 155)),
+            (None, True, [[0.1, 0.2, 0.3], [0.4, 0.5, 0.6]], (102, 102, 155, 155)),
+        ),
+    )
+    def it_can_locate_tiles(
+        self, request, tmpdir, alpha, pass_tiles, outline, expected_topleft
+    ):
+        slide, _ = base_test_slide(tmpdir, PILIMG.RGBA_COLOR_500X500_155_249_240)
+        _extract_tile = method_mock(request, Slide, "extract_tile")
+
+        coords = CP(0, 10, 0, 10)
+        tile = Tile(PILIMG.RGBA_COLOR_500X500_155_249_240, coords)
+        mock_tiles = [(tile, coords), (tile, coords)]
+
+        _tiles_generator = method_mock(request, GridTiler, "_tiles_generator")
+        _tiles_generator.return_value = mock_tiles
+
+        _has_enough_tissue = method_mock(request, Tile, "has_enough_tissue")
+        _has_enough_tissue.side_effect = True
+        _grid_coordinates_generator = method_mock(
+            request, GridTiler, "_grid_coordinates_generator"
+        )
+        _grid_coordinates_generator.return_value = [CP(0, 10, 0, 10), CP(0, 10, 0, 10)]
+        _extract_tile.side_effect = mock_tiles
+        grid_tiler = GridTiler((10, 10), level=0, check_tissue=True, tissue_percent=60)
+        binary_mask = BiggestTissueBoxMask()
+
+        if pass_tiles:
+            tiles = list(grid_tiler._tiles_generator(slide, binary_mask))
+        else:
+            tiles = None
+
+        img = grid_tiler.locate_tiles(
+            slide=slide,
+            extraction_mask=binary_mask,
+            alpha=alpha,
+            outline=outline,
+            tiles=tiles,
+        )
+        img = np.uint8(img)
+
+        assert img.shape == (15, 15, 4 if alpha else 3)
+        assert tuple(img[:4, 0, 0]) == expected_topleft
 
 
 class Describe_ScoreTiler:


alessiamarcolini:
Hey @kheffah thanks for this PR!! First of all, could you split this commit into 3 separate commits?

kheffah:
@alessiamarcolini All done :)

src/histolab/tiler.py
```diff
@@ -90,25 +92,39 @@ def locate_tiles(
             The alpha level to be applied to the rescaled slide, default to 128.
         outline: str
             The outline color for the tile annotations, default to 'red'.
+        linewidth: int
+            Thickness of line used to draw tiles
+        tiles: Iterable
+            Tiles to visualize. Optional, will be extracted if None.
```
ernestoarbitrio, line 98: I think this part of docstring is not clear. I would use something more verbose

src/histolab/tiler.py
```diff
@@ -90,25 +92,39 @@ def locate_tiles(
...
 
         Returns
         -------
         PIL.Image.Image
             PIL Image of the rescaled slide with the extracted tiles outlined
         """
         img = slide.scaled_image(scale_factor)
-        img.putalpha(alpha)
+        if alpha is not None:
```
ernestoarbitrio, line 106: Why alpha can be none if it has a default value? I think that alpha cannot be none and if a user passes none an error would be expected, that is fine I think

ernestoarbitrio (changes requested):
@kheffah thanks for this contribution. That's great. Just a couple of comments

src/histolab/tiler.py
kheffah: @ernestoarbitrio Actually this is on purpose. The default value is to fade out the slide image. But some users (myself included) prefer _not_ to fade out the background image, in which case they'd pass ``alpha=None``.  This is similar to how PIL.MAX_IMAGE_PIXELS has a default numerical value, but users can opt out by passing it as ``None``. Let me know if that makes sense to you :)

ernestoarbitrio: Ok, I see, but IMHO `alpha` has to be 255 that means no transparency and so the image will be without any fade i think. This way we could get rid of that `if` statement.

src/histolab/tiler.py
```diff
@@ -73,7 +73,9 @@ def locate_tiles(
         extraction_mask: BinaryMask = BiggestTissueBoxMask(),
         scale_factor: int = 32,
         alpha: int = 128,
-        outline: str = "red",
+        outline: Union[str, List] = "red",
+        linewidth: int = 1,
+        tiles: Iterable = None,
```
ernestoarbitrio, line 78: I'd like to see what kind of object for the iterable e.g. `Iterable[str]` this way is more meaningful when we read the code and we know that we're expecting an iterable of `Tile` objects or something like that. Hope I have been clear 🤷🏽‍♂️

src/histolab/tiler.py
```diff
@@ -90,25 +92,39 @@ def locate_tiles(
...
-        )
-        tiles_coords = (tile[1] for tile in tiles)
-        for coords in tiles_coords:
+        if tiles is None:
+            tiles = (
+                self._tiles_generator(slide, extraction_mask)[0]
+                if isinstance(self, ScoreTiler)
+                else self._tiles_generator(slide, extraction_mask)
+            )
+        tiles_coords = [tile[1] for tile in tiles]
```
ernestoarbitrio, line 116: i would keep it as a `tuple` as before:
`tiles_coords = (tile[1] for tile in tiles)`

src/histolab/tiler.py
```diff
@@ -90,25 +92,39 @@ def locate_tiles(
...
+
+        # maybe user specified colors for each tile
```
ernestoarbitrio, line 118: comments like this are not meaningful in my opinion. We should use comments for specify something that could be not clear in the future. So in this case the comment should be removed if the statement or the code portion is clear enough, otherwise the statement has to be improved XD

src/histolab/tiler.py
```diff
@@ -73,7 +73,9 @@ def locate_tiles(
         extraction_mask: BinaryMask = BiggestTissueBoxMask(),
         scale_factor: int = 32,
         alpha: int = 128,
-        outline: str = "red",
+        outline: Union[str, List] = "red",
```
ernestoarbitrio, line 76: I'm not sure this will work properly. We are using a union of string and list for the sake of the retro-compatibility but this can be a problem:
1. This implies a lot of if statement in the code that in turn requires a bunch of unit/integration tests with some edge cases.
2. the outline when it's a list and the length of the list is greater or less than the number of the tiles we will throw several `IndexError` or wrong color according to the size of the lists. See line `124`.

I wonder if we have to simplify this method or in some way introduce a validator for the outline.

If something is not clear please let me know 😄

tests/unit/test_tiler.py
```diff
@@ -777,6 +778,53 @@ def it_knows_whether_coordinates_are_within_extraction_mask(
...
         assert coords_within_extraction_mask == expected_result
 
+    @pytest.mark.parametrize(
+        "alpha, pass_tiles, outline, expected_topleft",
+        (
+            (128, False, "red", (255, 255, 155, 155)),
+            (None, True, [[0.1, 0.2, 0.3], [0.4, 0.5, 0.6]], (102, 102, 155, 155)),
+        ),
+    )
+    def it_can_locate_tiles(
```
ernestoarbitrio, line 788: I can buy this unit test, but given the several changes I'd like to see at least 1 integration test for the edge cases used here

ernestoarbitrio (changes requested):
Per comments, please let me know if something is not clear enough

src/histolab/tiler.py
kheffah: OK, all addressed.

src/histolab/tiler.py
kheffah: I agree. Now everything is a generator to preserve memory and time.

src/histolab/tiler.py
kheffah: That's a great point! I added a new generator that zips outline and tiles and throws errors for the edge cases. These edge cases were all added to unit tests.

tests/unit/test_tiler.py
kheffah: Actually, the edge cases are best resolved with a unit test. I added a unit test specifically for the edge cases. I also added integration test cases to make sure the new outline functionality produces the expected output.

ernestoarbitrio: ok nice

src/histolab/tiler.py
```diff
@@ -100,19 +122,62 @@ def locate_tiles(
...
+        if tiles is None:
+            tiles = (
+                self._tiles_generator(slide, extraction_mask)[0]
+                if isinstance(self, ScoreTiler)
+                else self._tiles_generator(slide, extraction_mask)
+            )
         tiles_coords = (tile[1] for tile in tiles)
-        for coords in tiles_coords:
+
+        for coords, one_outline in self._tile_coords_and_outline_generator(
```
ernestoarbitrio, line 133: I'm not sure `_tile_coords_and_outline_generator` is the right approach here. My general concern is that basically we are checking for the outline type and then we return the coord and the outline as they are passed to the method if they respect all the constraint. So it's just a validator, IMHO i do prefer to have a validator as you had in one of your pervious commit. Also i think that we could have loss of performance with a large number of Tiles and outlines 🤷🏽‍♂️

src/histolab/tiler.py
```diff
@@ -100,19 +122,62 @@ def locate_tiles(
...
+        ):
             rescaled = scale_coordinates(coords, slide.dimensions, img.size)
-            draw.rectangle(tuple(rescaled), outline=outline)
+            draw.rectangle(tuple(rescaled), outline=one_outline, width=linewidth)
         return img
 
     # ------- implementation helpers -------
 
+    @staticmethod
+    def _tile_coords_and_outline_generator(
```
ernestoarbitrio, line 143: FYI: We decided that all methods and properties (public or private) should be A-Z ordered so please keep in mind that all new method should be placed in the right position :D

ernestoarbitrio (changes requested):
I think we're close. Could you address my last comments please :D

tests/integration/test_tiler.py
```diff
@@ -18,13 +18,32 @@
 
 class DescribeRandomTiler:
     @pytest.mark.parametrize(
-        "fixture_slide, binary_mask, tile_size, level, check_tissue, expectation",
+        "fixture_slide, binary_mask, tile_size, level, outline, check_tissue, expectation",
```
ernestoarbitrio, line 21: this line is 91 char. Be sure you have black formatter before commit

src/histolab/tiler.py
kheffah: I beg to differ here. It's not just a validator .. depending on the type of ``outline`` argument, it either calls ``for coords in tiles_coords`` or ``for coords, one_outline in zip_longest(tiles_coords, outlines)``. So if the outline is a ``str`` it would return the same value for all coords, whereas if it is an ``Iterable`` it would fetch the value corresponding to the relevant tile and throw an exception is the lengths don't match. Even if I wanted to use a validator I couldn't because the number of tiles is not known beforehand in ``GridTiler`` so the only way to check the difference in length between it and the ``outline`` ``Iterable`` is using something like the ``zip_longest``.

src/histolab/tiler.py
kheffah: @ernestoarbitrio OK, addressed.

tests/integration/test_tiler.py
kheffah: @ernestoarbitrio Actually I did use black .. it automatically runs on all my commits with the precommit-hooks. I ran it again now and it didn't make any changes to that file. Are you sure 91 char is the maximum allowed?

ernestoarbitrio: 88 is the max

ernestoarbitrio:
```
.\tests\integration\test_tiler.py:21:89: E501 line too long (91 > 88 characters)
```

src/histolab/tiler.py
ernestoarbitrio: __gotcha__ ok

tests/integration/test_tiler.py
ernestoarbitrio: black doesn't take care of single strings > than 88 chars

kheffah: OK, fixed.

ernestoarbitrio (approved):

src/histolab/tiler.py
```diff
@@ -84,12 +87,31 @@ def locate_tiles(
...
             BinaryMask object defining how to compute a binary mask from a Slide.
             Default `BiggestTissueBoxMask`
-        scale_factor: int
+        scale_factor: int, optional
             Scaling factor for the returned image. Default is 32.
-        alpha: int
-            The alpha level to be applied to the rescaled slide, default to 128.
-        outline: str
-            The outline color for the tile annotations, default to 'red'.
+        alpha: int, optional
```
alessiamarcolini, line 92:
```suggestion
        alpha : int, optional
```

src/histolab/tiler.py
```diff
@@ -84,12 +87,31 @@ def locate_tiles(
...
+            The alpha level to be applied to the rescaled slide. Default is 128.
+        outline: Union[str, Iterable[str], Iterable[Tuple[int]]], optional
```
alessiamarcolini, line 94:
```suggestion
        outline : Union[str, Iterable[str], Iterable[Tuple[int]]], optional
```

src/histolab/tiler.py
```diff
@@ -84,12 +87,31 @@ def locate_tiles(
...
+            each color is your assigned color for the corresponding individual
+            tile. This list can be a list of matplotlib-style string colors, or
+            it maybe a list of tuples of ints in the [0, 255] range, each of
+            length 3, representing the red, green and blue color for each tile.
+            For example, if you have two tiles that you want to be colored
+            yellow, you can pass this argument as any of the following ..
+            - 'yellow'
+            - ['yellow', 'yellow']
+            - [(255, 255, 0), (255, 255, 0)]
+        linewidth: int, optional
```
alessiamarcolini, line 107:
```suggestion
        linewidth : int, optional
```

src/histolab/tiler.py
```diff
@@ -84,12 +87,31 @@ def locate_tiles(
...
+            Thickness of line used to draw tiles. Default is 1.
+        tiles: Iterable[Tile], optional
```
alessiamarcolini, line 109:
```suggestion
        tiles : Optional[Iterable[Tile]], optional
```

src/histolab/tiler.py
```diff
@@ -73,7 +74,9 @@ def locate_tiles(
         extraction_mask: BinaryMask = BiggestTissueBoxMask(),
         scale_factor: int = 32,
         alpha: int = 128,
-        outline: str = "red",
+        outline: Union[str, Iterable[str], Iterable[Tuple[int]]] = "red",
+        linewidth: int = 1,
+        tiles: Iterable[Tile] = None,
```
alessiamarcolini, line 79:
```suggestion
        tiles: Optional[Iterable[Tile]] = None,
```

src/histolab/tiler.py
```diff
@@ -84,12 +87,31 @@ def locate_tiles(
...
+            - 'yellow'
+            - ['yellow', 'yellow']
+            - [(255, 255, 0), (255, 255, 0)]
+        linewidth: int, optional
+            Thickness of line used to draw tiles. Default is 1.
+        tiles: Iterable[Tile], optional
+            Tiles to visualize. Will be extracted if None. Default is None.
+            You may decide to provide this argument if you do not want the
+            tiles to be re-extracted for visualization if you already have
+            the tiles in hand from, say, the ``_tiles_generator`` method of
```
alessiamarcolini, line 113: mmm why are we making a reference to a private method here..?

src/histolab/tiler.py
```diff
@@ -132,6 +158,45 @@ def _has_valid_tile_size(self, slide: Slide) -> bool:
...
+    @staticmethod
+    def _tile_coords_and_outline_generator(
+        tiles_coords: Iterable[CoordinatePair],
+        outlines: Union[str, List[str], List[Tuple[int]]],
+    ) -> Union[str, Tuple[int]]:
+        """Zip tile coordinates and outlines from tile and outline iterators.
+
+        Parameters
+        ----------
+        tiles_coords: Iterable[CoordinatePair]
```
alessiamarcolini, line 170:
```suggestion
        tiles_coords : Iterable[CoordinatePair]
```

src/histolab/tiler.py
```diff
@@ -132,6 +158,45 @@ def _has_valid_tile_size(self, slide: Slide) -> bool:
...
+            Coordinates referring to the upper left and lower right corners.
+        outlines: Union[str, Iterable[str], Iterable[Tuple[int]]]
```
alessiamarcolini, line 172:
```suggestion
        outlines : Union[str, Iterable[str], Iterable[Tuple[int]]]
```

src/histolab/tiler.py
```diff
@@ -132,6 +158,45 @@ def _has_valid_tile_size(self, slide: Slide) -> bool:
...
+    def _tile_coords_and_outline_generator(
+        tiles_coords: Iterable[CoordinatePair],
+        outlines: Union[str, List[str], List[Tuple[int]]],
+    ) -> Union[str, Tuple[int]]:
+        """Zip tile coordinates and outlines from tile and outline iterators.
+
+        Parameters
+        ----------
+        tiles_coords: Iterable[CoordinatePair]
+            Coordinates referring to the upper left and lower right corners.
```
alessiamarcolini, line 171:
```suggestion
            Coordinates referring to the tiles' upper left and lower right corners.
```

src/histolab/tiler.py
```diff
@@ -132,6 +158,45 @@ def _has_valid_tile_size(self, slide: Slide) -> bool:
...
+        outlines: Union[str, Iterable[str], Iterable[Tuple[int]]]
+            See docstring for ``locate_tiles`` for details.
+
+        Yields
+        -------
+        CoordinatePair
+            Coordinates referring to the upper left and lower right corners.
```
alessiamarcolini, line 178:
```suggestion
            Coordinates referring to the tiles' upper left and lower right corners.
```

src/histolab/tiler.py
```diff
@@ -132,6 +158,45 @@ def _has_valid_tile_size(self, slide: Slide) -> bool:
...
+        Union[str, Tuple[int]]
+            Fixed outline depending on user input to used by method ``locate_tiles``.
+        """
+        if isinstance(outlines, str):
+            for coords in tiles_coords:
+                yield coords, outlines
+
+        elif hasattr(outlines, "__iter__"):
+            for coords, one_outline in zip_longest(tiles_coords, outlines):
+                if any(j is None for j in (coords, one_outline)):
```
alessiamarcolini, line 188: can this be replaced with
`if None in (coords, one_outline)` ?

tests/unit/test_tiler.py
```diff
@@ -777,6 +778,98 @@ def it_knows_whether_coordinates_are_within_extraction_mask(
...
+                "str, Iterable[str], or Iterable[List[int]]",
+            ),
+        ),
+    )
+    def it_throws_error_with_invalid_tile_outline(
+        self, request, tmpdir, outline, n_tiles, error_msg
+    ):
+        slide, _ = base_test_slide(tmpdir, PILIMG.RGBA_COLOR_500X500_155_249_240)
+        _extract_tile = method_mock(request, Slide, "extract_tile")
+
```
alessiamarcolini, line 807:
```suggestion
```

tests/unit/test_tiler.py
```diff
@@ -777,6 +778,98 @@ def it_knows_whether_coordinates_are_within_extraction_mask(
...
+        coords = CP(0, 10, 0, 10)
+        tile = Tile(PILIMG.RGBA_COLOR_500X500_155_249_240, coords)
+        mock_tiles = [(tile, coords)] * n_tiles
+
```
alessiamarcolini, line 811:
```suggestion
```

tests/unit/test_tiler.py
```diff
@@ -777,6 +778,98 @@ def it_knows_whether_coordinates_are_within_extraction_mask(
...
+        _tiles_generator = method_mock(request, GridTiler, "_tiles_generator")
+        _tiles_generator.return_value = mock_tiles
+
```
alessiamarcolini, line 814:
```suggestion
```

tests/unit/test_tiler.py
```diff
@@ -777,6 +778,98 @@ def it_knows_whether_coordinates_are_within_extraction_mask(
...
+        _has_enough_tissue = method_mock(request, Tile, "has_enough_tissue")
+        _has_enough_tissue.side_effect = True
+        _grid_coordinates_generator = method_mock(
+            request, GridTiler, "_grid_coordinates_generator"
+        )
+        _grid_coordinates_generator.return_value = [CP(0, 10, 0, 10), CP(0, 10, 0, 10)]
+        _extract_tile.side_effect = mock_tiles
+        grid_tiler = GridTiler((10, 10))
```
alessiamarcolini, line 822:
```suggestion
        grid_tiler = GridTiler((10, 10))
        
```

tests/unit/test_tiler.py
```diff
@@ -777,6 +778,98 @@ def it_knows_whether_coordinates_are_within_extraction_mask(
...
+                "The parameter ``outline`` should be of type: "
+                "str, Iterable[str], or Iterable[List[int]]",
+            ),
+        ),
+    )
+    def it_throws_error_with_invalid_tile_outline(
+        self, request, tmpdir, outline, n_tiles, error_msg
+    ):
+        slide, _ = base_test_slide(tmpdir, PILIMG.RGBA_COLOR_500X500_155_249_240)
+        _extract_tile = method_mock(request, Slide, "extract_tile")
```
alessiamarcolini, line 806: I'm not sure I understand why we need this

tests/unit/test_tiler.py
```diff
@@ -777,6 +778,98 @@ def it_knows_whether_coordinates_are_within_extraction_mask(
...
+        _tiles_generator = method_mock(request, GridTiler, "_tiles_generator")
+        _tiles_generator.return_value = mock_tiles
+
+        _has_enough_tissue = method_mock(request, Tile, "has_enough_tissue")
+        _has_enough_tissue.side_effect = True
+        _grid_coordinates_generator = method_mock(
+            request, GridTiler, "_grid_coordinates_generator"
+        )
+        _grid_coordinates_generator.return_value = [CP(0, 10, 0, 10), CP(0, 10, 0, 10)]
+        _extract_tile.side_effect = mock_tiles
```
alessiamarcolini, line 821: same

tests/unit/test_tiler.py
```diff
@@ -777,6 +778,98 @@ def it_knows_whether_coordinates_are_within_extraction_mask(
...
+        binary_mask = BiggestTissueBoxMask()
+
+        img = grid_tiler.locate_tiles(
+            slide=slide,
+            extraction_mask=binary_mask,
+            alpha=alpha,
+            outline=outline,
+            tiles=None
+            if not pass_tiles
+            else list(grid_tiler._tiles_generator(slide, binary_mask)),
```
alessiamarcolini, line 866: if statements in test code are considered a bad practice, could you refactor this test to avoid this?
I think we already discussed this in a previous PR (now I don't remember...)

tests/unit/test_tiler.py
```diff
@@ -777,6 +778,98 @@ def it_knows_whether_coordinates_are_within_extraction_mask(
...
+        )
+        img = np.uint8(img)
+
+        assert img.shape == (15, 15, 4 if alpha else 3)
```
alessiamarcolini, line 870: in this case you can parametrize this to avoid the if statement

alessiamarcolini (changes requested):
some stylistic changes + unit test code shouldn't have conditionals (see comments)

tests/unit/test_tiler.py
ernestoarbitrio: alpha doesn't mean anything now cause it cannot be None anymore ... so i remove it

tests/unit/test_tiler.py
ernestoarbitrio: I think we can use this if to parametrize a test without duplicating all the stuff... If statements are bad in case of assertions but i think we can buy this

tests/unit/test_tiler.py
```diff
@@ -778,6 +779,73 @@ def it_knows_whether_coordinates_are_within_extraction_mask(
...
+        coords = CP(0, 10, 0, 10)
+        tile = Tile(PILIMG.RGBA_COLOR_500X500_155_249_240, coords)
+        mock_tiles = [(tile, coords), (tile, coords)]
+        method_mock(request, GridTiler, "_tiles_generator", return_value=mock_tiles)
+        method_mock(
+            request,
+            GridTiler,
+            "_grid_coordinates_generator",
+            return_value=[CP(0, 10, 0, 10), CP(0, 10, 0, 10)],
+        )
```
alessiamarcolini, line 840: why do we need this? this method is not called in the `locate_tiles` method, and we mock `_tiles_generator` that would call it

moreover if we mock something it's also better to assert that the mock has been called

alessiamarcolini (changes requested):

tests/unit/test_tiler.py
ernestoarbitrio: > why do we need this? this method is not called in the `locate_tiles` method, and we mock `_tiles_generator` that would call it
yes u right this is not needed
> 
> moreover if we mock something it's also better to assert that the mock has been called
that's not true Alessia, if we mock someting for the sake of isolation it's not mandatory to assert that something has been called ... we mock a lot of property for example and we do not assert anything 🤷🏽‍♂️

ernestoarbitrio (approved):

PR #308, opened 2021-08-07
Handle edge case for _are_coordinates_within_extraction_mask 

kheffah:
Handle edge case for ``_are_coordinates_within_extraction_mask`` where tile coordinates are off by 1 or 2 pixels due to conversion of floats to ints and usage of np.ceil or np.floor.

## Description

Only if the tile coordinates are valid does the function return True. Explicitly checking if ``tile_area`` > 0 avoids ``DivisionByZeroError``. This edge case is rare, but I've encountered it in some of the slides I'm working with.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/src/histolab/tiler.py
+++ b/src/histolab/tiler.py
@@ -333,7 +333,7 @@ class GridTiler(Tiler):
         tile_area = np.count_nonzero(tile_thumb_mask)
         tile_in_binary_mask_area = np.count_nonzero(tile_in_binary_mask)
 
-        return (
+        return tile_area > 0 and (
             tile_in_binary_mask_area / tile_area
             > COORDS_WITHIN_EXTRACTION_MASK_THRESHOLD
         )
--- a/tests/unit/test_tiler.py
+++ b/tests/unit/test_tiler.py
@@ -758,6 +758,7 @@ class Describe_GridTiler:
     @pytest.mark.parametrize(
         "tile_coords, expected_result",
         [
+            (CP(2, 6, 6, 7), False),  # bad edge coordinates from np.ceil/floor
             (CP(0, 0, 2, 2), False),  # completely outside of region
             (CP(0, 0, 8, 8), False),  # only 205
             (CP(2, 3, 6, 6), True),  # 85% in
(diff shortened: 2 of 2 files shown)


ernestoarbitrio (approved):

alessiamarcolini:
Hi @kheffah, I remember facing some 'ZeroDivisionError's when I was developing this method and somehow I managed to avoid them (at least with the slides I was working on). 

However, does it make sense a tile area of 0? Can we fix the cause instead of the symptom? 

Do you know by any chance why is only happening with some slides?

kheffah:
@alessiamarcolini Well I don't know every single case, but one of the ones I worked on had this problem for the a couple of tiles at the very edge of tissue, cause by approximation from ``float`` to ``int`` in [this line of code](https://github.com/histolab/histolab/blob/1ae4ce3c188d110a15fab9a2b8bf683efd55185a/src/histolab/util.py#L239). I think as long as tiling coordinates are extracted from a low resolution mask this will always be a potential issue due to scaling difference .. hence this PR.

kheffah:
@alessiamarcolini In some other work, I sometimes used the package ``shapely`` to check if some shape (eg square tile) fits within another (eg slide mask). ``Shapely`` does everything in the coordinate space, without creating any masks. I think doing this here is an overkill because: 1. It's an extra dependency for an uncommon edge case that we can easily handle otherwise, 2. It includes a lot of extra logic about polygons like how to account for holes etc.

alessiamarcolini:
> cause by approximation from float to int in this line of code. 

I understand

> I think as long as tiling coordinates are extracted from a low resolution mask this will always be a potential issue due to scaling difference 

Yes, but if we work with masks it is necessary to work with low resolution masks for RAM and time efficiency (we tried in the very early days...)

> Shapely does everything in the coordinate space, without creating any masks.

Very nice, we can take this into account for future developments, but I agree that for the moment it's an overkill

alessiamarcolini (approved):

Issue #311, opened 2021-08-18
`TissueMask` mask on `Tile`


alessiamarcolini:
**Describe the solution you'd like**
I would like to use the `TissueMask` mask also on a `Tile` instance, with the filters composition that histolab has for tiles

**Additional context**
It would be cool to use a type dispatcher on the `_mask` method to change the behaviour according to the type of the input

PR #313, opened 2021-08-18
Implement `TissueMask` mask for `Tile` with type dispatcher

alessiamarcolini:
## Description

Now is possible to use the `TissueMask` mask also on `Tile`s:

```python
from histolab.masks import TissueMask

tile = Tile(pil_img, None, None)
tissue_mask = TissueMask()

result = tissue_mask(tile) # same dimensions as the tile
```

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Fixes: #311 

## Checklist


ernestoarbitrio (changes requested):
please address the minor comment

src/histolab/masks.py
```diff
@@ -152,3 +155,23 @@ def _mask(self, slide) -> np.ndarray:
         filters = FiltersComposition(histolab.slide.Slide).tissue_mask_filters
         thumb_mask = filters(thumb)
         return thumb_mask
+
+    @lru_cache(maxsize=100)
+    @_mask.register(Tile)
+    def _(self, tile: Tile) -> np.ndarray:
+        """Return the thumbnail binary mask of the tissue area.
```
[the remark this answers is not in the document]
alessiamarcolini: fair enough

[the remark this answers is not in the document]
alessiamarcolini: I tried to redefine the `__call__` method in the `TissueMask` class, but I'm struggling to put type hints:

- if I do `from .slide import Slide` it fails due to circular imports
- with both`import histolab` and `import histolab.slide` I get `AttributeError: module 'histolab' has no attribute 'slide'`
- with `from histolab import slide` I would get a LGTM alert because the module is already imported as `import histolab`

I could just leave it without annotation and explain it only in the docstring, but I don't like it, any help @ernestoarbitrio ?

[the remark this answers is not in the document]
ernestoarbitrio: I think define `_` is ok, and i don't think it will be rendered in the docs ... anyhow the docstring should contain something meaningful for future development

alessiamarcolini:
hey @ernestoarbitrio what do I do with this CodeFactor issue?
```
src/histolab/filters/compositions.py#L91
Unnecessary pass statement (unnecessary-pass)
```

I wanted that method for the sake of the documentation, can I achieve the same result in another way?

src/histolab/filters/compositions.py
```diff
@@ -47,6 +55,49 @@ def __new__(cls: type, cls_):
             f"Filters composition for the class {cls_.__name__} is not available"
         )
 
+    @lazyproperty
+    def tissue_mask_filters(self) -> imf.Compose:
```
ernestoarbitrio, line 59: why we need this here? I don't see the reason of the empty method in here only for the doc. Maybe we should use a `raise NotImplementedError` in the super class otherwise it looks weird to me

ernestoarbitrio (changes requested):

src/histolab/filters/compositions.py
```diff
@@ -47,6 +55,50 @@ def __new__(cls: type, cls_):
...
+            - `image_filters.OtsuThreshold()
+              <image_filters.html#src.histolab.filters.image_filters.OtsuThreshold>`_
+
+            - `morphological_filters.BinaryDilation()
+              <morphological_filters.html#src.histolab.filters.morphological_filters.BinaryDilation>`_
+
+            - `morphological_filters.BinaryFillHoles(structure=np.ones((5, 5)))
+              <morphological_filters.html#src.histolab.filters.morphological_filters.BinaryFillHoles>`_
+        """
+        raise NotImplementedError
```
ernestoarbitrio, line 100: I would use a meaningful message for this exception

alessiamarcolini: mmmm, this error will never never never be raised, because the `__new__` instantiates another class

ernestoarbitrio: So we need it only for docstring?

alessiamarcolini: Yes exactly

ernestoarbitrio: mmmm I never faced something like this ... but it seems weird to me.

alessiamarcolini: Mmm I never used the factory pattern in Python before 😅

ernestoarbitrio: Actually even if the complexity is demanded to the factory ot other stupid methods, this docstring should be splitted into its derived methods in `TileComposition` and `SlideComposition` lazyproperties and get rid of this ugly thing

alessiamarcolini: Mmmm, but the classes are private and I guess the factory is used to hide the complexity.. how do we do this in practice?

ernestoarbitrio: oh ok you right now i got it. But i would use a message anyway like: 
```
        raise NotImplementedError(
            "must be implemented by each subclass"
        )  # pragma: no cover
        ```

alessiamarcolini: alright coming

ernestoarbitrio (approved):

Issue #309, opened 2021-08-10
Row-column X, Y coordinate mix-up in RandomTiler


kheffah:
**Describe the bug**
``RandomTiler`` and ``random_choice_mask2d`` sometimes mixes up row-column coordinates (eg. from numpy array shapes) with x-y coordinates for the actual slide.

**Software (please complete the following information):**
 - OS: Ubuntu18
 - Python Version: 3.8
 - histolab version: 0.2.6

**Additional context**

Where I saw this:

**random_choice_mask2d** 
[These lines](https://github.com/histolab/histolab/blob/d2a0bf09bde4b3e60e34dfa9ffbc8433d098f8c6/src/histolab/util.py#L105-L107) should probably be replaced by something like:
```python
    true_y, true_x = np.where(binary_mask)
    loc = np.random.randint(len(true_y) - 1)

    return true_x[loc], true_y[loc]
```
which is also more efficient because it calls ``np.where`` only once. Test expectations will need to be re-saved etc. Current version return row-column coordinates, not x-y.

**random_tile_coordinates**
This "compensates" for the issue above by passing x instead of y (and y instead of x) in [this line](https://github.com/histolab/histolab/blob/d2a0bf09bde4b3e60e34dfa9ffbc8433d098f8c6/src/histolab/tiler.py#L664). Of course, a double bug doesn't make things right as it makes maintenance hard and error prone. But putting that aside, this doesn't actually compensate, since the row coordinate gets added to the width, and column coordinate to the height, [here](https://github.com/histolab/histolab/blob/d2a0bf09bde4b3e60e34dfa9ffbc8433d098f8c6/src/histolab/tiler.py#L660-L661).

**How to fix**

``random_choice_mask2d`` should return X-Y coordinates .. using something like the code snippet above. ``random_tile_coordinates`` should be kept as-is **except** [this line](https://github.com/histolab/histolab/blob/d2a0bf09bde4b3e60e34dfa9ffbc8433d098f8c6/src/histolab/tiler.py#L664), where we should reverse X and Y.

PR #317, opened 2021-08-31
Fix row-column/X-Y coordinates mix-up in Random Tiler

alessiamarcolini:
## Description

This PR fixes the mismatch between row-column / X-Y coordinates in the `RandomTiler`. In the end, the result **was correct** due to a double bug, but know they should both be fixed.

Moreover, it fixes inverted `y_ul` and `x_br` in `CoordinatePair` used in unit testing (causing no harm), and it moves the conversion from np coords to PIL coords to `regions_from_binary_mask`

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes: #309

## Checklist


src/histolab/util.py
```diff
@@ -100,24 +100,24 @@ def random_choice_true_mask2d(binary_mask: np.ndarray) -> Tuple[int, int]:
     Returns
     -------
     Tuple[int, int]
-        Random pair of indices where the ``binary_mask`` is True.
+        Random pair of indices (column, row) where the ``binary_mask`` is True.
     """
-    x = np.random.choice(np.where(binary_mask)[0])
-    y = np.random.choice(np.where(binary_mask[x])[0])
+    true_y, true_x = np.where(binary_mask)
```
ernestoarbitrio, line 105: IMHO this should still be `x,y` the `true` prefix is weird to me. The fact the it calculates the pair of indices where the mask is true is specified well enough in the docstring, so this variables should maintain the old names

src/histolab/util.py
```diff
@@ -155,7 +155,11 @@ def regions_from_binary_mask(binary_mask: np.ndarray) -> List[Region]:
     thumb_labeled_regions = label(binary_mask)
     regions = [
         Region(
-            index=i, area=rp.area, bbox=rp.bbox, center=rp.centroid, coords=rp.coords
+            index=i,
+            area=rp.area,
+            bbox=rp.bbox[:2][::-1] + rp.bbox[2:][::-1],  # from np coords to PIL coords
```
ernestoarbitrio, line 160: can you explain this? It's too weird this statement 🤷🏽‍♂️

tests/integration/test_tiler.py
```diff
@@ -123,7 +123,7 @@ def it_locates_tiles_on_the_slide(
         slide = Slide(fixture_slide, "")
         random_tiles_extractor = RandomTiler(
             tile_size=tile_size,
-            n_tiles=2,
+            n_tiles=100,
```
ernestoarbitrio, line 126: did you consider the performance during testrunner increasing the n tiles of 2 order of magnitude?

tests/integration/test_util.py
```diff
@@ -17,8 +18,9 @@
     ],
 )
 def test_random_choice_true_mask2d_find_right_coordinates(fixture_slide, binary_mask):
+    np.random.seed(0)
```
ernestoarbitrio, line 21: why is this seed needed?

setup.cfg
```diff
@@ -2,7 +2,7 @@
 looponfailroots = tests
 python_classes = Test Describe
 python_files = test_*.py
-python_functions = test_ it_ they_ but_ and_it_ or_it
+python_functions = test_ it_ they_ but_ and_it_ or_it and_
```
ernestoarbitrio, line 5: I believe that the `and_` is not necessary. We should use the verbose `and_it` and so the new tests should be `and_it_does_not_xxxxxx`

tests/unit/test_tiler.py
```diff
@@ -672,7 +673,7 @@ def but_with_wrong_coordinates(self, request, tmpdir):
 
     def and_doesnt_raise_error_with_wrong_coordinates(self, request, tmpdir):
```
ernestoarbitrio, line 674: i know this is old ... but please replace the name with : `and_it_does_not_raise_error_with_wrong_coordinates` . In general tests name should be verbose and this name contains some weird thing like `doesnt` that it is not even english :D and the `it` is missing.

ernestoarbitrio (changes requested):
I like the fix ❤️  thx

Some comments before i can do another review round :D

tests/integration/test_util.py
alessiamarcolini: since we are testing a function with a random behaviour it's better for the test to be predictable

tests/integration/test_tiler.py
alessiamarcolini: I did consider it, but the point is that with only 2 tiles we cannot really see how the tiler is sampling the tiles, if you compare the old and new expectations you will see that only now this test is really meaninfgul

src/histolab/util.py
alessiamarcolini: basically, if we assume the original `rp.bbox` to be (1,2,3,4), the result of `rp.bbox[:2][::-1] + rp.bbox[2:][::-1]` would be  (2,1,4,3), so I'm flipping the order of the first two elements and then the order of the second two elements. All of this to convert from np coordinates to PIL coordinates

ernestoarbitrio (changes requested):

alessiamarcolini:
@ernestoarbitrio to do what you suggest I need to unroll the list comprehension to a normal for statement.. I will do that

ernestoarbitrio:
> @ernestoarbitrio to do what you suggest I need to unroll the list comprehension to a normal for statement.. I will do that

Actually i think you can still use comprehension if you write an internat function that given the `rp.bbox` it gives you back the correct new bbox shape. Hope i have been clear

alessiamarcolini:
actually this code is wrong.. 

> bbox = list(reversed(rp.bbox))
and then in the Region pass bbox=[*bbox[:2], *bbox[2:]]

I need to "split" the tuple in two and then individually reverse.. anyway I will implement it in a subfunction

ernestoarbitrio (approved):
LGTM, just 1 minor thing. Good to go after this

kheffah (approved):
Thanks @alessiamarcolini for this important PR! All looks good to me.

Issue #288, opened 2021-07-19
Stain normalization: Macenko


alessiamarcolini:
[paper](http://wwwx.cs.unc.edu/~mn/sites/default/files/macenko2009.pdf)

alessiamarcolini:
RGB to OD filter issue #289

PR #318, opened 2021-09-17
Add Stain Normalization with Macenko's method

alessiamarcolini:
## Description

Add Stain Normalization with Macenko's method

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Fixes: #288

## Checklist

## TODO before merge
- [x] finish unit tests
- [x] write integration tests
- [x] create API documentation page
- [ ] create documentation with examples/tutorial


src/histolab/mixins.py
```diff
@@ -0,0 +1,44 @@
...
+class LinalgMixin:
+    @staticmethod
+    def normalize_columns(arr: np.ndarray) -> np.ndarray:
+        """Normalize each column vector in an array"""
+        return arr / np.linalg.norm(arr, axis=0)
+
+    @staticmethod
+    def two_principal_components(arr: np.ndarray) -> np.ndarray:
+        """
+        Return the two principal components of the covariance matrix of ``arr``.
```
nicolebussola, line 31: the first two

src/histolab/filters/image_filters_functional.py
```diff
@@ -539,32 +540,34 @@ def rgb_to_lab(
     return lab
```
nicolebussola, line 540: return hsv

src/histolab/stain_normalizer.py
```diff
@@ -0,0 +1,299 @@
...
+        ----------
+        img_rgb : PIL.Image.Image
+            Image to normalize.
+        background_intensity : int, optional
+            Background transmitted light intensity. Default is 240.
+
+        Returns
+        -------
+        PIL.Image.Image
+            Stain normalized image
```
nicolebussola, line 82: Image with normalized stain

src/histolab/stain_normalizer.py
```diff
@@ -0,0 +1,299 @@
...
+
+        # determine concentrations of the individual stains
+        C = np.linalg.lstsq(stain_matrix, OD, rcond=None)[0]
+
+        return C
+
+
+class MacenkoStainNormalizer(LinalgMixin, TransformerStainMatrixMixin):
+
+    stain_color_map = {
```
nicolebussola, line 121: Here I would add a comment/reference

src/histolab/stain_normalizer.py
```diff
@@ -0,0 +1,299 @@
...
+        img_rgb, stain_matrix: np.ndarray, background_intensity: int = 240
+    ) -> np.ndarray:
+        OD = RgbToOd(background_intensity)(img_rgb)
+        # rows correspond to channels (RGB), columns to OD values
+        OD = np.reshape(OD, (-1, 3)).T
+
+        # determine concentrations of the individual stains
+        C = np.linalg.lstsq(stain_matrix, OD, rcond=None)[0]
+
+        return C
```
ernestoarbitrio, line 116: I would return directly `np.linalg.lstsq(stain_matrix, OD, rcond=None)[0]`, plus why that uppercase names for variables? Uppercase name are supposed to be used for constants

src/histolab/stain_normalizer.py
```diff
@@ -0,0 +1,299 @@
...
+            background_intensity, np.exp(-self.stain_matrix_target.dot(C2))
+        )
+        img_norm = np.clip(img_norm, a_min=None, a_max=255)
+        img_norm = np.reshape(img_norm.T, (*img_rgb.size[::-1], 3))
+        return np_to_pil(img_norm)
+
+    @staticmethod
+    def _find_concentrations(
+        img_rgb, stain_matrix: np.ndarray, background_intensity: int = 240
+    ) -> np.ndarray:
```
ernestoarbitrio, line 108: docstring is missed

src/histolab/stain_normalizer.py
```diff
@@ -0,0 +1,299 @@
...
+        OD = np.reshape(OD, (-1, 3)).T
+
+        # determine concentrations of the individual stains
+        C = np.linalg.lstsq(stain_matrix, OD, rcond=None)[0]
+
+        return C
+
+
+class MacenkoStainNormalizer(LinalgMixin, TransformerStainMatrixMixin):
+
```
ernestoarbitrio, line 120: docstring

src/histolab/stain_normalizer.py
```diff
@@ -0,0 +1,299 @@
...
+        # Convert to OD and ignore background
+        tile = Tile(img_rgb, None, None)
+        tissue_mask = TissueMask()
+        mask = tissue_mask(tile)
+
+        OD = RgbToOd(background_intensity)(img_rgb)
+        OD = OD[mask].reshape(-1, 3)
+
+        # Remove data with OD intensity less than β
+        ODhat = OD[~np.any(OD < beta, axis=1)]
```
ernestoarbitrio, line 188: same concern for variable names

src/histolab/filters/image_filters_functional.py
alessiamarcolini: sorry, why?

nicolebussola: ups, my bad

src/histolab/mixins.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class LinalgMixin:
+    @staticmethod
+    def normalize_columns(arr: np.ndarray) -> np.ndarray:
+        """Normalize each column vector in an array"""
+        return arr / np.linalg.norm(arr, axis=0)
+
+    @staticmethod
+    def two_principal_components(arr: np.ndarray) -> np.ndarray:
```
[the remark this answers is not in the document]
alessiamarcolini: it would make sense, but I cannot seem to make it work with the "canonical way of doing PCA with eigenvectors decomposition", something like this

```python
eigen_values, eigen_vectors = np.linalg.eigh(np.cov(arr, rowvar=False))
sorted_index = np.argsort(eigen_values)[::-1]
sorted_eigenvectors = eigen_vectors[:, sorted_index]

return sorted_eigenvectors[:, 0:2]
```

[staintools](https://github.com/Peter554/StainTools/blob/master/staintools/stain_extraction/macenko_stain_extractor.py#L33) and [torchstain](https://github.com/EIDOSlab/torchstain/blob/main/torchstain/normalizers/numpy_macenko_normalizer.py#L29) are just getting the two middle columns,
while [histomicstk](https://github.com/DigitalSlideArchive/HistomicsTK/blob/588ecc7b96a5fc2a41b88f80115c780ab26e4edc/histomicstk/preprocessing/color_deconvolution/_linalg.py#L9) is using SVD, and for that specific input they don't have to slice anything


so... honestly I just took what worked and in this situation I don't know how to generalize this

src/histolab/stain_normalizer.py
```diff
@@ -0,0 +1,299 @@
...
+            if the input image is not RGB or RGBA
+
+        References
+        ----------
+        .. [1] Macenko, Marc, et al. "A method for normalizing histology slides for
+            quantitative analysis." 2009 IEEE International Symposium on Biomedical
+            Imaging: From Nano to Macro. IEEE, 2009.
+        """
+
+        if stains is not None and len(stains) != 2:
```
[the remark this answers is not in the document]
alessiamarcolini: very good question.. 😅 I looked around and I couldn't find a proper answer
@kheffah why is histomicstk only supporting a two-stains list?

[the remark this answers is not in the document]
nicolebussola: I do not think it is such a big problem, hardly slides present more than 2 stains, it was a curiosity; wanted to know what is the problem with more stains

src/histolab/mixins.py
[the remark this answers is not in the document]
alessiamarcolini: yes we are slicing but not how one would expect 😅 donno

src/histolab/filters/image_filters.py
```diff
@@ -286,10 +286,12 @@ class RgbToOd(ImageFilter):
...
     img : PIL.Image.Image
         Input image
+    background_intensity : int, optional
+        Background transmitted light intensity. Default is 240.
 
     Returns
     -------
-    PIL.Image.Image
+    np.ndarray
         Image in OD space
```
ernestoarbitrio, line 295: this is not an image anymore, is a numpy array

src/histolab/mixins.py
```diff
@@ -0,0 +1,44 @@
...
+        arr : np.ndarray
+            Input array
+
+        Returns
+        -------
+        np.ndarray
+            Two principal components.
+        """
+        _, V = np.linalg.eigh(np.cov(arr, rowvar=False))
+        return V[:, 1:3]
```
ernestoarbitrio, line 44: we should call it `principal_components` and pass a parameters `start` and  `end` that ins this case is `1:3` what do you think?

ernestoarbitrio (changes requested):
couple of comments (minor)

src/histolab/mixins.py
nicolebussola: yes, just give me a little time to look into this

src/histolab/mixins.py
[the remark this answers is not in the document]
nicolebussola: in practice, to compute the PCs there are several methods, including using the covariance matrix or the singular value decomposition (svd):

- case 1 i.e. we use the covariance matrix: it turns out that ```np.linalg.eigh``` returns the eigenvalues  in order of increasing magnitude, thus the two principal eigenvectors are the last ones, meaning that either we sort the columns and take the first 2 or we take the last two (as done by staintools). 

- case 2  i.e. we use the svd approach: in this case, we should first transpose our arr and then apply the svd, this would give us directly the PC matrix, but we have to slide it anyway (as done by [histomicstk](https://github.com/DigitalSlideArchive/HistomicsTK/blob/7860903b44dd564190d2234e475b7516af136127/histomicstk/preprocessing/color_deconvolution/separate_stains_macenko_pca.py#L85) ) to take the first two.

I think the second approach is preferable, as we are working with 2-stain images we do not need more components, however we can parametrize this as suggested by @ernestoarbitrio

Issue #294, opened 2021-07-21
Add conversion level-magnification 


nicolebussola:
### Discussed in https://github.com/histolab/histolab/discussions/293

<div type='discussions-op-text'>

<sup>Originally posted by **afvrbanac** July 20, 2021</sup>
Hello, novice user here with some basic questions. I'm using histolab to extract random tiles at different levels. Is it possible to automatically add a scale bar to these images? 

Also, is the magnification of each level the original magnification divided by the downsample for the level? So for example, for a 40X image at level 0, if the downsample for level 2 was 16, the magnification would be 2.5X?</div>

PR #319, opened 2021-09-20
add conversion level - magnification factor, when possible

nicolebussola:
Address Issue #294

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -115,6 +115,27 @@ dimensions at native magnification or at a specified level:
    Dimensions at level 1: (7500, 8496)
    Dimensions at level 2: (1875, 2124)
 
+.. note::
+    If the native magnification, *i.e.* the magnification factor used to scan the slide, is provided in the slide properties, it is also possible
+    to convert the desired level to its corresponding magnification factor with the ``level_magnification_factor`` property.
+
+.. code-block:: ipython3
+
+   print(
+        "Native magnification factor:",
+        prostate_slide.level_magnification_factor()
+    )
+
+    print(
+        "Magnification factor corresponding to level 1:",
+        prostate_slide.level_magnification_factor(level=1),
+    )
+
+.. code-block:: bash
+
+    Native magnification factor: 20X
+    Magnification factor corresponding to level 1: 5.0X
+
 Moreover, we can retrieve or show the slide thumbnail in a separate window:
 
 .. code-block:: ipython3
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -116,7 +116,7 @@ dimensions at native magnification or at a specified level:
    Dimensions at level 2: (1875, 2124)
 
 .. note::
-    If the native magnification, *i.e.* the magnification factor used to scan the slide, is provided in the slide properties, it is also possible
+    If the native magnification, *i.e.*, the magnification factor used to scan the slide, is provided in the slide properties, it is also possible
     to convert the desired level to its corresponding magnification factor with the ``level_magnification_factor`` property.
 
 .. code-block:: ipython3
--- a/src/histolab/exceptions.py
+++ b/src/histolab/exceptions.py
@@ -37,6 +37,10 @@ class LevelError(HistolabException):
     """Raised when a requested level is not available"""
 
 
+class PropertyError(HistolabException):
+    """Raised when a requested slide property is not available"""
+
+
 class FilterCompositionError(HistolabException):
     """Raised when a filter composition for the class is not available"""
 
--- a/src/histolab/exceptions.py
+++ b/src/histolab/exceptions.py
@@ -37,7 +37,7 @@ class LevelError(HistolabException):
     """Raised when a requested level is not available"""
 
 
-class PropertyError(HistolabException):
+class SlidePropertyError(HistolabException):
     """Raised when a requested slide property is not available"""
 
 
--- a/src/histolab/slide.py
+++ b/src/histolab/slide.py
@@ -26,7 +26,7 @@ import openslide
 import PIL
 from skimage.measure import find_contours
 
-from .exceptions import LevelError, PropertyError
+from .exceptions import LevelError, SlidePropertyError
 from .filters.compositions import FiltersComposition
 from .tile import Tile
 from .types import CoordinatePair
@@ -48,6 +48,11 @@ class Slide:
         Path where the WSI is saved.
     processed_path : Union[str, pathlib.Path]
         Path where the tiles will be saved to.
+
+    Raises
+    ------
+    TypeError:
+        If the processed path is not specified.
     """
 
     def __init__(
@@ -125,6 +130,11 @@ class Slide:
         -------
         dimensions : Tuple[int, int]
             Slide dimensions at the specified level (width, height)
+
+        Raises
+        ------
+        IndexError:
+            If the specified level is not available
         """
         level = level if level >= 0 else self._remap_level(level)
         try:
@@ -137,6 +147,7 @@ class Slide:
 
     def level_magnification_factor(self, level: int = 0) -> str:
         """Return the magnification factor at the specified level.
+
          Notice that the conversion level-magnification can be computed only
         if the native magnification is available in the slide metadata.
 
@@ -149,6 +160,14 @@ class Slide:
         -------
         magnification factor : str
             Magnification factor at speficied level
+
+        Raises
+        ------
+        LevelError:
+            If the specified level is not available.
+        KeyError:
+            If the slide's native magnification is not available in the file
+            metadata.
         """
         level = level if level >= 0 else self._remap_level(level)
         if level > (len(self._wsi.level_dimensions) + 1):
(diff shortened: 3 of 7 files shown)


src/histolab/exceptions.py
```diff
@@ -37,6 +37,10 @@ class LevelError(HistolabException):
     """Raised when a requested level is not available"""
 
 
+class PropertyError(HistolabException):
```
ernestoarbitrio, line 40: I would call it with the `Slide` prefix considering that the docstring says :`Raised when a requested slide property is not available` . So if it is a specific Exception for a slide I would use that prefix

`SlidePropertyError`

src/histolab/slide.py
```diff
@@ -135,6 +135,39 @@ def level_dimensions(self, level: int = 0) -> Tuple[int, int]:
...
+        Parameters
+        ---------
+        level : int
+            The level which magnification factor is requested, default is 0
+
+        Returns
+        -------
+        magnification factor : str
+            Magnification factor at speficied level
+        """
```
ernestoarbitrio, line 152: Miss the 2 Raises in the docstring

ernestoarbitrio (changes requested):
per comments

docs/quickstart.rst
```diff
@@ -114,6 +114,27 @@ dimensions at native magnification or at a specified level:
    Dimensions at level 1: (7500, 8496)
    Dimensions at level 2: (1875, 2124)
 
+.. note::
+    If the native magnification, *i.e.* the magnification factor used to scan the slide, is provided in the slide properties, it is also possible
```
alessiamarcolini, line 118:
```suggestion
    If the native magnification, *i.e.,* the magnification factor used to scan the slide, is provided in the slide properties, it is also possible
```

src/histolab/slide.py
```diff
@@ -48,6 +48,11 @@ class Slide:
         Path where the WSI is saved.
     processed_path : Union[str, pathlib.Path]
         Path where the tiles will be saved to.
+
+    Raises
+    ------
+    TypeError:
```
alessiamarcolini, line 54:
```suggestion
    TypeError
```

src/histolab/slide.py
```diff
@@ -125,6 +130,11 @@ def level_dimensions(self, level: int = 0) -> Tuple[int, int]:
         -------
         dimensions : Tuple[int, int]
             Slide dimensions at the specified level (width, height)
+
+        Raises
+        ------
+        IndexError:
```
alessiamarcolini, line 136:
```suggestion
        LevelError
```

src/histolab/slide.py
```diff
@@ -135,6 +145,53 @@ def level_dimensions(self, level: int = 0) -> Tuple[int, int]:
...
+    def level_magnification_factor(self, level: int = 0) -> str:
+        """Return the magnification factor at the specified level.
+
+         Notice that the conversion level-magnification can be computed only
+        if the native magnification is available in the slide metadata.
+
+        Parameters
+        ---------
+        level : int
+            The level which magnification factor is requested, default is 0
```
alessiamarcolini, line 157:
```suggestion
            The level which magnification factor is requested, default is 0.
```

src/histolab/slide.py
```diff
@@ -135,6 +145,53 @@ def level_dimensions(self, level: int = 0) -> Tuple[int, int]:
...
+
+        Returns
+        -------
+        magnification factor : str
+            Magnification factor at speficied level
+
+        Raises
+        ------
+        LevelError:
```
alessiamarcolini, line 166:
```suggestion
        LevelError
```

src/histolab/slide.py
```diff
@@ -135,6 +145,53 @@ def level_dimensions(self, level: int = 0) -> Tuple[int, int]:
...
+            If the specified level is not available.
+        KeyError:
```
alessiamarcolini, line 168:
```suggestion
        SlidePropertyError
```

src/histolab/slide.py
```diff
@@ -135,6 +145,53 @@ def level_dimensions(self, level: int = 0) -> Tuple[int, int]:
...
+            If the slide's native magnification is not available in the file
+            metadata.
+        """
+        level = level if level >= 0 else self._remap_level(level)
+        if level > (len(self._wsi.level_dimensions) + 1):
+            raise LevelError(
+                f"Level {level} not available. Number of available levels: "
+                f"{len(self._wsi.level_dimensions)}"
+            )
+        properties = self._wsi.properties
```
alessiamarcolini, line 178: I would use our wrapper
```suggestion
        properties = self.properties
```

src/histolab/slide.py
```diff
@@ -135,6 +145,53 @@ def level_dimensions(self, level: int = 0) -> Tuple[int, int]:
...
+        Raises
+        ------
+        LevelError:
+            If the specified level is not available.
+        KeyError:
+            If the slide's native magnification is not available in the file
+            metadata.
+        """
+        level = level if level >= 0 else self._remap_level(level)
+        if level > (len(self._wsi.level_dimensions) + 1):
```
alessiamarcolini, line 173:
```suggestion
        if level in self.levels:
```

src/histolab/slide.py
```diff
@@ -135,6 +145,53 @@ def level_dimensions(self, level: int = 0) -> Tuple[int, int]:
...
+            else 1
+        )
+        try:
+            level_magnification = (
+                int(properties["openslide.objective-power"]) / downsample_factor
+            )
+            return f"{level_magnification}X"
+        except KeyError:
+            raise SlidePropertyError(
+                f"Native magnification not available. Available slide properties: "
```
alessiamarcolini, line 191:
```suggestion
                "Native magnification not available. Available slide properties: "
```

src/histolab/slide.py
```diff
@@ -135,6 +145,53 @@ def level_dimensions(self, level: int = 0) -> Tuple[int, int]:
...
+                f"{list(self._wsi.properties.keys())}"
```
alessiamarcolini, line 192:
```suggestion
                f"{list(self.properties.keys())}"
```

tests/unit/test_slide.py
```diff
@@ -277,6 +278,80 @@ def but_it_raises_expection_when_level_does_not_exist(self, level, tmpdir):
...
+
+        with pytest.raises(LevelError) as err:
+            slide.level_magnification_factor(level=level)
+
+        assert isinstance(err.value, LevelError)
+        assert (
+            str(err.value)
+            == f"Level {level} not available. Number of available levels: 3"
+        )
+
```
alessiamarcolini, line 354: Usually we test exceptions with unit tests to make them fast (mocking where needed)

tests/unit/test_slide.py
```diff
@@ -277,6 +278,80 @@ def but_it_raises_expection_when_level_does_not_exist(self, level, tmpdir):
...
+                marks=pytest.mark.skipif(not on_ci(), reason="To run only on CI"),
+            ),
+        ),
+    )
+    def it_knows_its_magnification_factors(self, fixture_slide, level, expected_value):
+        slide = Slide(fixture_slide, "")
+
+        magnification_factor = slide.level_magnification_factor(level=level)
+
+        assert magnification_factor == expected_value
```
alessiamarcolini, line 309: these tests should go with the integration tests as we are using real slides, and unit tests for this method are missing :)

alessiamarcolini (changes requested):

docs/quickstart.rst
nicolebussola: not sure its needed

alessiamarcolini: it depends whether we are writing British or American. I guess we always used American english

alessiamarcolini (changes requested):
Now the tests that are included are ok. 

However, since the integration test `it_knows_its_magnification_factors` is basically run only on CI, I would add:
* a test case for that method to use the local `cmu-1-small-region.svs`
* two unit tests using a fake svs and mocking `properties` property to include or not `"openslide.objective-power"` and `"f"openslide.level[{level}].downsample"` to better control the flow. Let's try to have 100% coverage on the new modifications without the help of the integration tests

src/histolab/slide.py
```diff
@@ -135,6 +145,59 @@ def level_dimensions(self, level: int = 0) -> Tuple[int, int]:
...
+            If the slide's native magnification is not available in the file
+            metadata.
+        """
+        level = level if level >= 0 else self._remap_level(level)
+        properties = self.properties
+
+        if level not in self.levels:
+            raise LevelError(
+                f"Level {level} not available. Number of available levels: "
+                f"{len(self._wsi.level_dimensions)}"
```
alessiamarcolini, line 178: here you should also use `len(self.levels)`, and then you will see that the unit test `and_it_raises_an_exception_if_level_in_incorrect` will fail (as it should)

tests/unit/test_slide.py
```diff
@@ -277,6 +277,109 @@ def but_it_raises_expection_when_level_does_not_exist(self, level, tmpdir):
...
+            (
+                {
+                    "openslide.objective-power": "20",
+                    "openslide.level[1].downsample": "4.003",
+                    "openslide.level[2].downsample": "16",
+                },
+            ),
+        ),
+    )
+    def and_it_raises_an_exception_if_level_in_incorrect(
```
alessiamarcolini, line 344:
```suggestion
    def and_it_raises_an_exception_if_level_is_incorrect(
```

tests/unit/test_slide.py
```diff
@@ -277,6 +277,109 @@ def but_it_raises_expection_when_level_does_not_exist(self, level, tmpdir):
...
+        self, request, properties, tmpdir
+    ):
+        slide, _ = base_test_slide(tmpdir, PILIMG.RGBA_COLOR_500X500_155_249_240)
+        property_mock(request, Slide, "properties", return_value=properties)
+        property_mock(request, Slide, "levels", return_value=[0, 1, 2])
+        with pytest.raises(LevelError) as err:
+            slide.level_magnification_factor(4)
+
+        assert isinstance(err.value, LevelError)
+        assert str(err.value) == "Level 4 not available. Number of available levels: 1"
```
alessiamarcolini, line 354: getting back to the previous comment... here `levels` is mocked as `[0, 1, 2]`, while the error message says that there is 1 level.... obv wrong :D

alessiamarcolini (changes requested):

alessiamarcolini (approved):

src/histolab/slide.py
```diff
@@ -135,6 +145,59 @@ def level_dimensions(self, level: int = 0) -> Tuple[int, int]:
...
+        magnification factor : str
+            Magnification factor at speficied level
+
+        Raises
+        ------
+        LevelError
+            If the specified level is not available.
+        SlidePropertyError
+            If the slide's native magnification is not available in the file
+            metadata.
```
ernestoarbitrio, line 170: I think here we should add an `or` in the sense that we raise that execption in 2 cases and we mentioned only 1. So I would include 
```
or if Downsample factor for given level is not available.
```

ernestoarbitrio (approved):
LGTM but let's see if you can address the minor comment I just wrote

ernestoarbitrio (approved):
LGTM well done 👯

Issue #322, opened 2021-09-28
Stain Normalization: Reinhard


alessiamarcolini:
[Paper](https://ieeexplore.ieee.org/document/946629)

PR #323, opened 2021-09-28
Add filter LabToRgb e fix return image in RgbToLab

alessiamarcolini:
## Description

In order to implement #322, we need to convert back from the LAB space to the RGB space.

Moreover, it fixes a bug in the filter `RgbToLab`, where by mistake the resulting Lab image was converted to a PIL image, messing up the value range and type (it should be float in the range −128 to 127).

## Types of Changes

- [ ] Core
- [x] Bugfix
- [x] New feature
- [ ] Documentation

* Fixes: 

## Checklist


src/histolab/filters/image_filters.py
```diff
@@ -253,7 +253,7 @@ class RgbToLab(ImageFilter):
 
     Returns
     -------
-    PIL.Image.Image
+    np.ndarray
```
ernestoarbitrio, line 256: this is not an Image anymore ... please change the docstring

src/histolab/filters/image_filters_functional.py
```diff
@@ -517,13 +543,14 @@ def rgb_to_lab(
...
+        The name of the illuminant (the function is NOT case sensitive). Default is
+        "D65".
     observer : {“2”, “10”}, optional
-        The aperture angle of the observer.
+        The aperture angle of the observer. Default is "2".
 
     Returns
     -------
-    PIL.Image.Image
+    np.ndarray
```
ernestoarbitrio, line 553: not an Image anymore

ernestoarbitrio (changes requested):
2 easy comments ,... than LGTM

alessiamarcolini:
PR #334 needs to be merged first

.github/workflows/tests.yml
```diff
@@ -29,6 +29,7 @@ jobs:
   build:
     needs: wsi-artifacts
     runs-on: ${{ matrix.os }}
+    continue-on-error: true
```
ernestoarbitrio, line 32: is this needed? I don't see why we should continue the CI if we do not have these fixtures

ernestoarbitrio (changes requested):
1 single comment then LGTM

.github/workflows/tests.yml
alessiamarcolini: yes you're right, leftovers

ernestoarbitrio (approved):

Issue #269, opened 2021-05-19
Getting issue with TCGA 


mobilist:
Hi Team,
getting issue with fetching the file from tcga repository https://portal.gdc.cancer.gov/repository?facetTab=files

inline below is the error log from jupyter notebook for your reference

Downloading file 'tcga/prostate/6b725022-f1d5-4672-8c6c-de8140345210' from 'https://api.gdc.cancer.gov/data/6b725022-f1d5-4672-8c6c-de8140345210' to '/root/.cache/histolab-images/0.2.5'.
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
/usr/local/lib/python3.8/dist-packages/histolab/data/__init__.py in _fetch(data_filename)
    189     try:
--> 190         resolved_path = image_fetcher.fetch(data_filename)
    191     except HTTPError as httperror:

/usr/local/lib/python3.8/dist-packages/pooch/core.py in fetch(self, fname, processor, downloader)
    538 
--> 539             stream_download(
    540                 url,

/usr/local/lib/python3.8/dist-packages/pooch/core.py in stream_download(url, fname, known_hash, downloader, pooch, retry_if_failed)
    723             with temporary_file(path=str(fname.parent)) as tmp:
--> 724                 downloader(url, tmp, pooch)
    725                 hash_matches(tmp, known_hash, strict=True, source=str(fname.name))

/usr/local/lib/python3.8/dist-packages/pooch/downloaders.py in __call__(self, url, output_file, pooch)
    179             response = requests.get(url, **kwargs)
--> 180             response.raise_for_status()
    181             content = response.iter_content(chunk_size=self.chunk_size)

/usr/local/lib/python3.8/dist-packages/requests/models.py in raise_for_status(self)
    942         if http_error_msg:
--> 943             raise HTTPError(http_error_msg, response=self)
    944 

HTTPError: 451 Client Error: UNAVAILABLE FOR LEGAL REASONS for url: https://api.gdc.cancer.gov/data/6b725022-f1d5-4672-8c6c-de8140345210

During handling of the above exception, another exception occurred:

HTTPError                                 Traceback (most recent call last)
<ipython-input-5-a20de3808a6e> in <module>
----> 1 prostate_svs, prostate_path = prostate_tissue()
      2 ovarian_svs, ovarian_path = ovarian_tissue()

/usr/local/lib/python3.8/dist-packages/histolab/data/__init__.py in prostate_tissue()
    476         Path where the slide is saved
[...]

PR #346, opened 2021-11-09
Map TCGA old UUID with new UUID to fix `data` module

alessiamarcolini:
## Description

Finally, the TCGA portal allows for downloading the slides, but with the new release they changed all the UUIDs of the files. 

This PR also saves the files in the cache with its original TCGA name instead of the UUID, so it can be easier to recover from if there are other changes from the TCGA side in the future.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [x] Documentation

* Fixes: #269

## Checklist

- [x] I have updated the documentation accordingly.


alessiamarcolini:
Good news is that for the TCGA example in `examples/TCGA`, the file with metadata `clinical_csv_example.csv` did not include UUIDs of files but only the `case_id`, which remained the same between the two releases, so there is no need to update it

ernestoarbitrio (approved):
❤️

nicolebussola (approved):
LGTM

Issue #357, opened 2021-11-25
`Tile.has_enough_tissue` sometimes returns `False` on tiles with tissue only


lorenz-gorini:
**Describe the bug**
`Tile.has_enough_tissue` method sometimes returns `False` on tiles with tissue only.
Sometimes the filters used to extract the tissue mask of the tile are not identifying the whole tissue, but the cell nuclei only. 
Therefore during tile extraction from the WSI, the tile will be wrongly discarded as if it contained too much background.

**To Reproduce**
Steps to reproduce the behavior (assuming that the attached `.png` files are saved in local path `~/Downloads`):
```python
>>> from histolab.tile import Tile
... from histolab.types import CoordinatePair
... tile = Tile("~/Downloads/tile_1_issue.png", CoordinatePair(3118, 2283, 3246, 2411))
... tile.has_enough_tissue(80, 0.1)
False
```
When inspecting  we found that ![`tile._has_tissue_more_than_percent(tissue_percent=80)`](https://github.com/histolab/histolab/blob/348f9451780bae10fdfcd85e05b05455106c896e/histolab/tile.py#L122) returns `False`.
To be more specific, we computed the tissue mask (that are used to check the tissue coverage) for some tiles by using the same code inside `Tile._has_tissue_more_than_percent(80)` method:

```python
>>> from histolab.filters.compositions import FiltersComposition
... filters = FiltersComposition(Tile).tissue_mask_filters
... img = Image.fromarray(filters(tile._image))
... img.save("tile_1_issue_mask.png")
```
and we found the attached masks `tile_1_issue_mask.png` and `tile_2_issue_mask.png`. 

**Expected behavior**
We expected that the tissue masks of the tiles (`tile_1_issue_mask.png` and `tile_2_issue_mask.png`) were covering the whole tile, so that  `tile._has_tissue_more_than_percent(tissue_percent=80)` would return True as well as `tile.has_enough_tissue(80, 0.1)`

**Screenshots**
![tile_1_issue](https://user-images.githubusercontent.com/53576429/143477117-187c02ab-0a30-4167-9512-41d10ae0c38b.png)
![tile_1_issue_mask](https://user-images.githubusercontent.com/53576429/143477122-eaf5550e-9591-4a83-9bd4-8f1f65bc1c2f.png)
![tile_2_issue](https://user-images.githubusercontent.com/53576429/143477123-9c63900b-7426-49a6-a783-06f4ccb66a5b.png)
[...]

PR #360, opened 2021-11-29
Fix `Tile`'s tissue mask in case of a full tissue tile

alessiamarcolini:
## Description

This PR fixes the calculation of the tissue mask (and related usages) in case we have a tile completely covered with tissue.

From #357 "_Sometimes the filters used to extract the tissue mask of the tile are not identifying the whole tissue, but the cell nuclei only. Therefore during tile extraction from the WSI, the tile will be wrongly discarded as if it contained too much background_."

In order to solve this, the filters are applied on the tile with a white border of 10px, and then the border is removed from the returned mask.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes: #357

## Checklist


histolab/tile.py
```diff
@@ -166,6 +166,53 @@ def save(self, path: Union[str, bytes, os.PathLike]) -> None:
...
+        # apply filters on tile with border
+        filters = FiltersComposition(Tile).tissue_mask_filters
+        mask_border = filters(tile_border)
+
+        # remove border
+        mask = mask_border[
+            border_thickness : np_tile.shape[0] + border_thickness,
+            border_thickness : np_tile.shape[1] + border_thickness,
+        ]
+        return mask
```
ernestoarbitrio, line 214: General concern of this property is the massive usage of variables. Considering that this is a lazyproperty and it will be cached we are using a lot of memory for storing memory.
For example:
`tile_border = PIL.Image.fromarray(container)` is really needed?
IMHO we should try to optimize the memory of this property in order to keep it as small as possible.

histolab/tile.py
```diff
@@ -166,6 +166,53 @@ def save(self, path: Union[str, bytes, os.PathLike]) -> None:
...
+        In order to address an issue with the calculation of the tissue mask in tiles
+        where the tissue covers all the tile area, the filters are applied on the tile
+        with a white border of 10px, and then the border is removed from the returned
+        mask.
+        """
+        np_tile = np.array(self._image)
+
+        border_thickness = 10
+
+        # create a "white" array 20px bigger than the tile
```
ernestoarbitrio, line 189: in the docstring there's a 10px reference but here I see 20px. I'm confuse 😄

histolab/tile.py
```diff
@@ -166,6 +166,53 @@ def save(self, path: Union[str, bytes, os.PathLike]) -> None:
         Path(path).parent.mkdir(parents=True, exist_ok=True)
         self._image.save(path)
 
+    @lazyproperty
+    def tissue_mask(self) -> np.ndarray:
+        """Return a binary mask representing the tissue in the tile.
```
ernestoarbitrio, line 171: This is a property that represents a thing, so I'd get rid of `Return`.

```
**Binary mask representing the tissue in the tile.**
```

ernestoarbitrio (changes requested):

histolab/tile.py
alessiamarcolini: `lru_cache` works in a way that it saves in a dictionary **only** the result of the called function, so honestly I wouldn't be concerned with the amount of variables we use internally (I can agree that it could be "heavy" the first time).

That considered, it is a "big" function, doing a task that I don't know how to reasonably break into sub-tasks, and my main focus when writing it was that it would be easy to read and understand. What do you think @ernestoarbitrio ?

histolab/tile.py
alessiamarcolini: it's a 10px border on each X and Y side, so in total the container is 20px bigger on the X and Y axis

histolab/tile.py
ernestoarbitrio: I'm not saying I would split into little pieces. I'm just saying to check if we can get rid of some assigment to free-up memory. Think of a pipeline where histolab is used. If this property is called in a foorloop or basically is called 1000 times we have to be sure we are doing memory de-allocation. What do you think?

histolab/masks.py
```diff
@@ -197,7 +197,4 @@ def _(self, tile: Tile) -> np.ndarray:
         mask: np.ndarray
```
nicolebussola, line 197: docstring.

Is it necessary to have this as a method?

histolab/tile.py
alessiamarcolini: To answer some of your questions in a data driven-way, I rescued to the use of a memory profiler.

**This is the implementation with reduced variables assignments**:
```python
    @lazyproperty
    def tissue_mask_less_assignments(self) -> np.ndarray:
        """Binary mask representing the tissue in the tile.

        Returns
        -------
        np.ndarray
            Binary mask representing the tissue in the tile.

        Notes
        -----
        In order to address an issue with the calculation of the tissue mask in tiles
        where the tissue covers all the tile area, the filters are applied on the tile
        with a white border of 10px, and then the border is removed from the returned
        mask.
        """
        np_tile = np.array(self._image)

        border_thickness = 10

        # create a "white" array 20px bigger than the tile
        container_size = ()
        container = (
            np.ones(
                (
                    np_tile.shape[0] + 2 * border_thickness,
                    np_tile.shape[1] + 2 * border_thickness,
                    np_tile.shape[2],
                ),
                dtype=np.uint8,
            )
            * 255
        )

        # paste the tile at the center
        container[
            border_thickness : np_tile.shape[0] + border_thickness,
            border_thickness : np_tile.shape[1] + border_thickness,
            :,
        ] = np_tile

        # apply filters on tile with border
        filters = FiltersComposition(Tile).tissue_mask_filters
        mask_border = filters(PIL.Image.fromarray(container))

        # remove border
        return mask_border[
            border_thickness : np_tile.shape[0] + border_thickness,
            border_thickness : np_tile.shape[1] + border_thickness,
        ]
```

**This is the code I used to profile the current implementation**:
```python
from memory_profiler import profile

from PIL import Image

img = Image.open("tests/fixtures/tiles/high-nuclei-score-level2.png")

tile = Tile(img, None, None)


@profile
def normal():
    a = tile.tissue_mask
    b = tile.tissue_mask
    c = tile.tissue_mask
    d = tile.tissue_mask
    e = tile.tissue_mask
    f = tile.tissue_mask
    g = tile.tissue_mask


normal()
```

**And this is the code I used to profile the modified implementation**:
```python
from memory_profiler import profile

from PIL import Image

img = Image.open("tests/fixtures/tiles/high-nuclei-score-level2.png")

tile = Tile(img, None, None)


@profile
def less_assignments():
    a = tile.tissue_mask_less_assignments
    b = tile.tissue_mask_less_assignments
    c = tile.tissue_mask_less_assignments
    d = tile.tissue_mask_less_assignments
    e = tile.tissue_mask_less_assignments
    f = tile.tissue_mask_less_assignments
    g = tile.tissue_mask_less_assignments


less_assignments()
```
## Results with 512x512 tile (140KB):

**Current**:
```Line #    Mem usage    Increment  Occurences   Line Contents
============================================================
    11     98.4 MiB     98.4 MiB           1   @profile
    12                                         def normal():
    13    106.9 MiB      8.6 MiB           1       a = tile.tissue_mask
    14    106.9 MiB      0.0 MiB           1       b = tile.tissue_mask
    15    106.9 MiB      0.0 MiB           1       c = tile.tissue_mask
    16    106.9 MiB      0.0 MiB           1       d = tile.tissue_mask
    17    106.9 MiB      0.0 MiB           1       e = tile.tissue_mask
    18    106.9 MiB      0.0 MiB           1       f = tile.tissue_mask
    19    106.9 MiB      0.0 MiB           1       g = tile.tissue_mask
```

**Modified**:
```
Line #    Mem usage    Increment  Occurences   Line Contents
============================================================
    11     98.4 MiB     98.4 MiB           1   @profile
    12                                         def less_assignments():
    13    107.0 MiB      8.6 MiB           1       a = tile.tissue_mask_less_assignments
    14    107.0 MiB      0.0 MiB           1       b = tile.tissue_mask_less_assignments
    15    107.0 MiB      0.0 MiB           1       c = tile.tissue_mask_less_assignments
    16    107.0 MiB      0.0 MiB           1       d = tile.tissue_mask_less_assignments
    17    107.0 MiB      0.0 MiB           1       e = tile.tissue_mask_less_assignments
    18    107.0 MiB      0.0 MiB           1       f = tile.tissue_mask_less_assignments
    19    107.0 MiB      0.0 MiB           1       g = tile.tissue_mask_less_assignments
```

## Results with 956 × 779 tile (1.7MB):

**Current**:
```
Line #    Mem usage    Increment  Occurences   Line Contents
============================================================
    11     98.5 MiB     98.5 MiB           1   @profile
    12                                         def normal():
    13    114.6 MiB     16.1 MiB           1       a = tile.tissue_mask
    14    114.6 MiB      0.0 MiB           1       b = tile.tissue_mask
    15    114.6 MiB      0.0 MiB           1       c = tile.tissue_mask
    16    114.6 MiB      0.0 MiB           1       d = tile.tissue_mask
    17    114.6 MiB      0.0 MiB           1       e = tile.tissue_mask
    18    114.6 MiB      0.0 MiB           1       f = tile.tissue_mask
    19    114.6 MiB      0.0 MiB           1       g = tile.tissue_mask
```

**Modified**:
```
Line #    Mem usage    Increment  Occurences   Line Contents
============================================================
    11     98.4 MiB     98.4 MiB           1   @profile
    12                                         def less_assignments():
    13    114.5 MiB     16.1 MiB           1       a = tile.tissue_mask_less_assignments
    14    114.5 MiB      0.0 MiB           1       b = tile.tissue_mask_less_assignments
    15    114.5 MiB      0.0 MiB           1       c = tile.tissue_mask_less_assignments
    16    114.5 MiB      0.0 MiB           1       d = tile.tissue_mask_less_assignments
    17    114.5 MiB      0.0 MiB           1       e = tile.tissue_mask_less_assignments
    18    114.5 MiB      0.0 MiB           1       f = tile.tissue_mask_less_assignments
    19    114.5 MiB      0.0 MiB           1       g = tile.tissue_mask_less_assignments
```

## Conclusions
The important bits of the profiler results are the increments on the first line/first call. They are exactly the same between the two implementations with a smaller and a bigger tile. Therefore I don't see the point in using fewer variables (which would just damage the readability)

ernestoarbitrio: awesome great answer

ernestoarbitrio (approved):

histolab/masks.py
alessiamarcolini: > docstring. 

what do you mean?

> Is it necessary to have this as a method?

I like the idea of having `TissueMask` that is flexible and does the right thing based on the input

@ernestoarbitrio thoughts?

ernestoarbitrio: I do not understand the matter ... sorry

Issue #375, opened 2022-01-17
Allow custom filters for TissueMask


kishoresurendra:
**Describe the bug**

The `RemoveSmallObjects()` class is removing essential small tissue islands in our WSI due to the pre-set filters.

In the current implementation, every operation under `FilterComposition`, i.e, `BinaryDilation()`, `RemoveSmallHoles()` and `RemoveSmallObjects()` have pre-determined filter sizes.

1.  For example, `RemoveSmallHoles()` has an `area_threshold` of 3000, but we may want a different threshold of, say 2000.  
2. Another example: `RemoveSmallObjects()` has a default `min_size` of 3000, but we may want to change it to 1500.

**To Reproduce**

1. `from histolab.masks import TissueMask`
2. `all_tissue_mask = TissueMask()`
3. `slidetox = Slide(slidepath, outputpath)`
4. `slidetox.locate_mask(all_tissue_mask).show()`

**Expected behavior**

To be able to add the custom filter(s) in the constructor of the `TissueMask`. In addition to the default constructor `TissueMask()`, we would like the following option as well:

```
TissueMask( 
    imf.RgbToGrayscale(),
    imf.OtsuThreshold(),
    mof.BinaryDilation(),
    mof.RemoveSmallObjects(min_size=2000),
    mof.RemoveSmallHoles(area_threshold=1500)
 )
```

or any other similar approach.

We are keen to provide the pool request for this issue.

ernestoarbitrio:
Hi @kishoresurendra thanks for reporting this! We're looking forward for you pull request 😄 . If you need anything else please let us know.

PR #377, opened 2022-01-25
Allow custom filters in TissueMask

Christopher22:
## Description

This small patch allows a custom "pipeline" of image filters within the TissueMask to optimize for specific demands regarding tissue. The default behavior is not changed.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes: #375 

## Checklist

### DIFF
--- a/histolab/filters/compositions.py
+++ b/histolab/filters/compositions.py
@@ -16,6 +16,8 @@
 # limitations under the License.
 # ------------------------------------------------------------------------
 
+from typing import Type
+
 import numpy as np
 
 from ..exceptions import FilterCompositionError
@@ -29,9 +31,10 @@ class FiltersComposition:
 
     Arguments
     ---------
-    cls_ : type, {Tile, Slide}
+    cls_ : type, {Tile, Slide, Compose}
         The class to get the appropriate filters composition for
-
+    *custom_filters : imf.ImageFilter
+        Custom filter applied if (and only if) the type Compose is used.
 
     Example:
         >>> from histolab.filters.compositions import FiltersComposition
@@ -41,12 +44,13 @@ class FiltersComposition:
         >>> filters_tile = FiltersComposition(Tile).tissue_mask_filters
     """
 
-    def __new__(cls: type, cls_):
+    def __new__(cls, cls_: type, *custom_filters: imf.ImageFilter):
         if not cls_:
             raise FilterCompositionError("cls_ parameter cannot be None")
         FiltersSubCls = {
             "Tile": _TileFiltersComposition,
             "Slide": _SlideFiltersComposition,
+            "Compose": _CustomFiltersComposition,
         }.get(cls_.__name__)
         if FiltersSubCls:
             instance = super(FiltersComposition, FiltersSubCls).__new__(FiltersSubCls)
@@ -139,3 +143,32 @@ class _TileFiltersComposition(FiltersComposition):
                 mof.BinaryFillHoles(structure=np.ones((20, 20))),
             ]
         )
+
+
+class _CustomFiltersComposition(FiltersComposition):
+    def __init__(
+        self, cls_: Type[imf.Compose], *custom_filters: imf.ImageFilter
+    ) -> None:
+        """Create a custom filter composition with at least one filter.
+
+        Generally, their is no need to call this constructor manually.
+        Please use ``FiltersComposition(Compose, *filters)`` instead.
+        """
+
+        if len(custom_filters) == 0:
+            raise FilterCompositionError(
+                "Custom filter pipeline requires at least one filter"
+            )
+
+        self._custom_filters = custom_filters
+
+    @lazyproperty
+    def tissue_mask_filters(self) -> imf.Compose:
+        """Filters composition for tissue estimation with custom filters.
+
+        Returns
+        -------
+        imf.Compose
+            Filters composition
+        """
+        return imf.Compose(self._custom_filters)
--- a/histolab/masks.py
+++ b/histolab/masks.py
@@ -24,6 +24,7 @@ import numpy as np
 import histolab
 
 from .filters.compositions import FiltersComposition
+from .filters.image_filters import Compose, ImageFilter
 from .slide import Slide
 from .tile import Tile
 from .types import Region
@@ -132,10 +133,22 @@ class BiggestTissueBoxMask(BinaryMask):
 
 
 class TissueMask(BinaryMask):
-    """Object that represent the whole tissue area mask.
+    """Object that represent the whole tissue area mask."""
 
-    The tissue within the slide or tile is automatically detected through a predefined
-    chain of filters."""
+    def __init__(self, *filters: ImageFilter) -> None:
+        """
+        Create a new tissue mask.
+
+        If custom image filters are specified, those are used instead the default ones.
+        By default, the tissue within the slide or tile is automatically detected
+        through a predefined chain of filters.
+
+        Parameters
+        ----------
+        *filters : ImageFilter
+            Custom filters to derive a TissueMask which overwrite the default pipeline.
+        """
+        self.custom_filters = filters
 
     def __call__(self, obj: Union[Slide, Tile]) -> np.ndarray:
         """Apply a predefined chain of filters to calculate the tissue area mask.
@@ -178,8 +191,14 @@ class TissueMask(BinaryMask):
             Binary mask of the tissue area. The dimensions are those of the thumbnail.
         """
         thumb = slide.thumbnail
-        filters = FiltersComposition(histolab.slide.Slide).tissue_mask_filters
-        thumb_mask = filters(thumb)
+
+        # Generate appropriate composition of filter
+        if len(self.custom_filters) == 0:
+            composition = FiltersComposition(Slide)
+        else:
+            composition = FiltersComposition(Compose, *self.custom_filters)
+
+        thumb_mask = composition.tissue_mask_filters(thumb)
         return thumb_mask
 
     @lru_cache(maxsize=100)
@@ -197,4 +216,11 @@ class TissueMask(BinaryMask):
         mask: np.ndarray
             Binary mask of the tissue area. The dimensions are those of the tile.
         """
-        return tile.tissue_mask
+
+        # Check if calculating a customized mask is required.
+        # Otherwise, fall back to the default one.
+        if len(self.custom_filters) > 0:
+            custom_filters = FiltersComposition(Compose, *self.custom_filters)
+            return tile.calculate_tissue_mask(custom_filters)
+        else:
+            return tile.tissue_mask
--- a/histolab/masks.py
+++ b/histolab/masks.py
@@ -222,5 +222,4 @@ class TissueMask(BinaryMask):
         if len(self.custom_filters) > 0:
             custom_filters = FiltersComposition(Compose, *self.custom_filters)
             return tile.calculate_tissue_mask(custom_filters)
-        else:
-            return tile.tissue_mask
+        return tile.tissue_mask
--- a/histolab/tile.py
+++ b/histolab/tile.py
@@ -170,6 +170,26 @@ class Tile:
     def tissue_mask(self) -> np.ndarray:
         """Binary mask representing the tissue in the tile.
 
+        It is calculated given a composition of suitable filters.
+        If a non-default composition is required, consider ``calculate_tissue_mask``.
+
+        Returns
+        -------
+        np.ndarray
+            Binary mask representing the tissue in the tile.
+        """
+        return self.calculate_tissue_mask(FiltersComposition(Tile))
+
+    def calculate_tissue_mask(
+        self, filter_composition: FiltersComposition
+    ) -> np.ndarray:
+        """Calculate the binary mask representing the tissue in the tile.
+
+        Parameters
+        ---------
+        filter_composition: FiltersComposition
+            FiltersComposition used to calculate the mask.
+
         Returns
         -------
         np.ndarray
@@ -200,7 +220,7 @@ class Tile:
         tile_border = PIL.Image.fromarray(np_tile_border)
 
         # apply filters on tile with border
-        filters = FiltersComposition(Tile).tissue_mask_filters
+        filters = filter_composition.tissue_mask_filters
         mask_border = filters(tile_border)
 
         # remove border
--- a/tests/unit/test_compositions.py
+++ b/tests/unit/test_compositions.py
@@ -14,6 +14,10 @@ from histolab.tile import Tile
 from ..unitutil import class_mock, initializer_mock
 
 
+class NotAllowedInFiltersComposition:
+    """A class NOT allowed to be used with FiltersComposition."""
+
+
 def it_knows_tissue_areas_mask_tile_filters_composition(
     RgbToGrayscale_, OtsuThreshold_, BinaryDilation_, BinaryFillHoles_
 ):
@@ -62,6 +66,29 @@ def it_knows_tissue_areas_mask_slide_filters_composition(
     assert type(_enough_tissue_mask_filters_) == Compose
 
 
+def it_knows_tissue_areas_mask_compose_filters_composition(
+    RgbToGrayscale_,
+    OtsuThreshold_,
+    BinaryDilation_,
+    RemoveSmallHoles_,
+    RemoveSmallObjects_,
+):
+    filters = [
+        RgbToGrayscale_(),
+        OtsuThreshold_(),
+        BinaryDilation_(),
+        RemoveSmallHoles_(),
+        RemoveSmallObjects_(),
+    ]
+
+    _enough_tissue_mask_filters_ = FiltersComposition(
+        Compose, *filters
+    ).tissue_mask_filters
+
+    assert type(_enough_tissue_mask_filters_) == Compose
+    assert list(_enough_tissue_mask_filters_.filters) == filters
+
+
 @pytest.mark.parametrize(
     "_cls, subclass",
     ((Tile, _TileFiltersComposition), (Slide, _SlideFiltersComposition)),
@@ -77,7 +104,7 @@ def it_can_dispatch_subclass_according_class_type(request, _cls, subclass):
 
 def it_raises_filtercompositionerror_if_class_not_allowed(request):
     _init_ = initializer_mock(request, FiltersComposition)
-    cls_ = Compose
+    cls_ = NotAllowedInFiltersComposition
 
     with pytest.raises(FilterCompositionError) as err:
         FiltersComposition(cls_)
@@ -85,7 +112,9 @@ def it_raises_filtercompositionerror_if_class_not_allowed(request):
     _init_.assert_not_called()
     assert isinstance(err.value, FilterCompositionError)
     assert (
-        str(err.value) == "Filters composition for the class Compose is not available"
+        str(err.value)
+        == "Filters composition for the class NotAllowedInFiltersComposition \
+is not available"
     )
 
 
@@ -99,6 +128,20 @@ def it_raises_filtercompositionerror_if_class_is_none(request):
     _init_.assert_not_called()
     assert isinstance(err.value, FilterCompositionError)
     assert str(err.value) == "cls_ parameter cannot be None"
+
+
+def it_raises_filtercompositionerror_if_compose_is_specified_without_filters(
+    request,
+):
+    _init_ = initializer_mock(request, FiltersComposition)
+    cls_ = Compose
+
+    with pytest.raises(FilterCompositionError) as err:
+        FiltersComposition(cls_)
+
+    _init_.assert_not_called()
+    assert isinstance(err.value, FilterCompositionError)
+    assert str(err.value) == "Custom filter pipeline requires at least one filter"
 
 
 # fixture components ---------------------------------------------
--- a/tests/unit/test_masks.py
+++ b/tests/unit/test_masks.py
@@ -1,8 +1,9 @@
 import numpy as np
+import PIL
 import pytest
 
 from histolab.filters.compositions import _SlideFiltersComposition
-from histolab.filters.image_filters import Compose
+from histolab.filters.image_filters import Compose, ImageFilter
 from histolab.filters.morphological_filters import RemoveSmallObjects
 from histolab.masks import BiggestTissueBoxMask, TissueMask
 from histolab.tile import Tile
@@ -16,6 +17,11 @@ from ..unitutil import (
     method_mock,
     property_mock,
 )
+
+
+class CustomFilterForTest(ImageFilter):
+    def __call__(self, _: PIL.Image.Image) -> np.ndarray:
+        raise RuntimeError("This call needs mocking")
 
 
 class DescribeBiggestTissueBoxMask:
@@ -133,10 +139,26 @@ class DescribeTissueMask:
             ]
         )
         tissue_mask = TissueMask()
-
         binary_mask = tissue_mask(slide)
 
         assert binary_mask == expected_mask
+
+    def it_knows_its_mask_slide_and_supports_custom_filters(self, request, tmpdir):
+        slide, _ = base_test_slide(tmpdir, PILIMG.RGBA_COLOR_500X500_155_249_240)
+        custom_filters = [CustomFilterForTest()]
+        custom_filter_call_ = method_mock(request, CustomFilterForTest, "__call__")
+        expected_mask = [
+            [True, True],
+            [False, True],
+        ]
+        custom_filter_call_.return_value = expected_mask
+
+        tissue_mask = TissueMask(*custom_filters)
+        binary_mask = tissue_mask(slide)
+
+        assert binary_mask == expected_mask
+        # Check the filter has been called once
+        custom_filter_call_.assert_called_once_with(custom_filters[0], slide.thumbnail)
 
     def it_knows_its_mask_tile(self, request):
         tile = Tile(PILIMG.RGBA_COLOR_500X500_155_249_240, None, None)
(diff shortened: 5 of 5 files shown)


tests/unit/test_masks.py
```diff
@@ -153,6 +154,17 @@ def it_knows_its_mask_tile(self, request):
...
+    def it_knows_custom_image_filters(self):
+        tile = Tile(PILIMG.RGBA_COLOR_500X500_155_249_240, None, None)
+
+        # Create a tissue mask which should not cover anything
+        tissue_mask = TissueMask(
+            image_filters.RgbToGrayscale(), image_filters.OtsuThreshold()
+        )
+
+        binary_mask = tissue_mask(tile)
+        assert not np.any(binary_mask)
```
ernestoarbitrio, line 166: This test smells to me. Here you are not testing the functionality. A unit test need to mock the objects involved in the actual implementation, here you are asserting something that does not imply nothing.

ernestoarbitrio (changes requested):
Thanks for your contribution @Christopher22 ❤️ !
I have some concerns on this PR 🤷🏽‍♂️ . The first thing is the tests you provided. Please read my comment. 
Furthermore @alessiamarcolini can give you other consideration on you work.

tests/unit/test_masks.py
Christopher22: Thank you for the review! To be honest, I could not completely get your point... With the unit test above, I just check that the result after applying the custom pipeline differs from the default (I might do that more explicit, as I rely on the current documentation there...) Testing the filters themselves is already covered in the other tests, is it not?

ernestoarbitrio: My point is that a unit tests that has to proof that your changes are working. So maybe I should've been more clear, sorry. Given that I'd like to see a unit tests that mocks your objects you introduced. For example how do we know that `_CustomFiltersComposition` has been called properly? Ho can we know that the `setter` within that is doing the right thing? Is it working only on a `Slide` or on a `Tile` too?

My hunch is that the changes of the custom filters works only if applied on a `Slide` and not on a `Tile` I'd like to see a tests with this kind of proof.

IMHO i do not think that the test `it_knows_custom_image_filters` is not applying the filters you believe. So to disprove my hypothesis I'd like to see a tests with this kind of assertions.

Hope I've been clear 🤷🏽‍♂️ . Let me know 😄

Christopher22: You are of cause entirely right! I will go through the code again. The missing implementation for Tile is almost embarrassing ... 

Is there any documentation regarding your preferred unit tests and mock functions? They are amazingly sophisticated

ernestoarbitrio: 👍🏽 thanks a lot. Anyway there's no official preference doc for unit testing. I'd suggest a quick reading of https://realpython.com/python-mock-library/. Furthermore if you want you can join the slack channel (link in the README) and we can discuss there... or even we can have a brief call and we can do the implementation together. Let me know what do you think

Christopher22: Thank you very much for the additional information! I will join the Slack channel which seems like a valuable resource. 

I used my spare time on the train and added both the missing implementation and (hopefully) proper unit tests. 

What do you think?

histolab/filters/compositions.py
```diff
@@ -139,3 +143,23 @@ def tissue_mask_filters(self) -> imf.Compose:
...
             ]
         )
+
+
+class _CustomFiltersComposition(FiltersComposition):
+    def __init__(self, *custom_filters) -> None:
+        super().__init__()
+        self.custom_filters = custom_filters
+
+    def set_custom_filters(self, *custom_filters):
```
ernestoarbitrio, line 153: Please could you add the missing return type and the type for the `custom_filters`

histolab/filters/compositions.py
```diff
@@ -139,3 +143,23 @@ def tissue_mask_filters(self) -> imf.Compose:
                 mof.BinaryFillHoles(structure=np.ones((20, 20))),
             ]
         )
+
+
+class _CustomFiltersComposition(FiltersComposition):
+    def __init__(self, *custom_filters) -> None:
```
ernestoarbitrio, line 149: Please could you add the type for `custom_filters`?

histolab/filters/compositions.py
```diff
@@ -41,15 +41,19 @@ class FiltersComposition:
         >>> filters_tile = FiltersComposition(Tile).tissue_mask_filters
     """
 
-    def __new__(cls: type, cls_):
+    def __new__(cls: type, cls_, *custom_filters):
```
ernestoarbitrio, line 44: Could you provide the missing typing annotation here?

histolab/filters/compositions.py
```diff
@@ -139,3 +143,23 @@ def tissue_mask_filters(self) -> imf.Compose:
...
+        self.custom_filters = custom_filters
+
+    def set_custom_filters(self, *custom_filters):
+        """To set the custom filters to be composed"""
+        self.custom_filters = custom_filters
+
+    @lazyproperty
+    def tissue_mask_filters(self) -> imf.Compose:
+        """Filters composition for slide's tissue estimation with custom filters
+        Returns
```
ernestoarbitrio, line 160:
```suggestion

        Returns
```

histolab/filters/compositions.py
```diff
@@ -139,3 +143,23 @@ def tissue_mask_filters(self) -> imf.Compose:
...
+        super().__init__()
+        self.custom_filters = custom_filters
+
+    def set_custom_filters(self, *custom_filters):
+        """To set the custom filters to be composed"""
+        self.custom_filters = custom_filters
+
+    @lazyproperty
+    def tissue_mask_filters(self) -> imf.Compose:
+        """Filters composition for slide's tissue estimation with custom filters
```
ernestoarbitrio, line 159:
```suggestion
        """Filters composition for tissue estimation with custom filters.
```

histolab/masks.py
```diff
@@ -134,8 +135,17 @@ def _regions(regions: List[Region], n: int = 1) -> List[Region]:
 class TissueMask(BinaryMask):
     """Object that represent the whole tissue area mask.
 
-    The tissue within the slide or tile is automatically detected through a predefined
-    chain of filters."""
+    By default, the tissue within the slide or tile is automatically
+    detected through a predefined chain of filters.
+    However, custom filters could be used, too.
```
ernestoarbitrio, line 140: I would move this part of the docstring from the class and place it within the `__init__`. Furthermore I'd get rid of the last line `However ... `

histolab/masks.py
```diff
@@ -22,6 +22,7 @@
 import numpy as np
 
 import histolab
+from histolab.filters import image_filters
```
ernestoarbitrio, line 25:
```suggestion
from .filters.image_filters import ImageFilter, MorphologicalFilter
```

histolab/masks.py
```diff
@@ -134,8 +135,17 @@ def _regions(regions: List[Region], n: int = 1) -> List[Region]:
...
     """Object that represent the whole tissue area mask.
 
-    The tissue within the slide or tile is automatically detected through a predefined
-    chain of filters."""
+    By default, the tissue within the slide or tile is automatically
+    detected through a predefined chain of filters.
+    However, custom filters could be used, too.
+    """
+
+    def __init__(self, *filters: image_filters.ImageFilter):
```
ernestoarbitrio, line 143:
```suggestion
    def __init__(self, *filters: List[Union[ImageFilter, MorphologicalFilter]]) -> None:
```

histolab/masks.py
```diff
@@ -134,8 +135,17 @@ def _regions(regions: List[Region], n: int = 1) -> List[Region]:
...
+        """
+        Create a new tissue mask. If custom image filters are specified,
+        those are used instead of the default ones.
```
ernestoarbitrio, line 146: IMHO considering the https://www.python.org/dev/peps/pep-0257/ the first line of the docstring should be a sort of summary and fit in 88 chars. All the other info can be underneath the summary line with a black line between them.

Moreover the `Parameters` section of the docstring is missed 😄

histolab/masks.py
```diff
@@ -178,7 +188,15 @@ def _mask(self, slide) -> np.ndarray:
             Binary mask of the tissue area. The dimensions are those of the thumbnail.
         """
         thumb = slide.thumbnail
-        filters = FiltersComposition(histolab.slide.Slide).tissue_mask_filters
+
+        # Generate appropiate composition of filter
```
ernestoarbitrio, line 192:
```suggestion
        # Generate appropriate composition of filter
```

histolab/masks.py
```diff
@@ -178,7 +188,15 @@ def _mask(self, slide) -> np.ndarray:
...
+        filters = (
+            FiltersComposition(histolab.slide.Slide)
+            if len(self.custom_filters) == 0
+            else FiltersComposition(
+                histolab.filters.image_filters.Compose, *self.custom_filters
+            )
+        ).tissue_mask_filters
```
ernestoarbitrio, line 199: Sorry but I see some problems in this portion of code:
1. `histolab.filters.image_filters.Compose` 👉🏽 could you import `Compose` without the whole reference of the histolab path?
2. The statement is too much convoluted maybe you should split it in something more readable (remember that the zen of python says: `beautiful is better than ugly 😂 `

histolab/masks.py
```diff
@@ -197,4 +215,13 @@ def _(self, tile: Tile) -> np.ndarray:
         mask: np.ndarray
             Binary mask of the tissue area. The dimensions are those of the tile.
         """
-        return tile.tissue_mask
+
+        # Check if calculating a customized mask is required.
+        # Otherwise, fall back to the default one.
+        if len(self.custom_filters) > 0:
+            custom_filters = FiltersComposition(
+                histolab.filters.image_filters.Compose, *self.custom_filters
```
ernestoarbitrio, line 223: As before 😄 ... import `Compose` and avoid the usage of the whole path.

histolab/tile.py
```diff
@@ -168,7 +168,22 @@ def save(self, path: Union[str, bytes, os.PathLike]) -> None:
 
     @lazyproperty
     def tissue_mask(self) -> np.ndarray:
-        """Binary mask representing the tissue in the tile.
+        """Binary mask representing the tissue in the tile calculated
```
ernestoarbitrio, line 171: please keep the first line and the body separated with a blank line, and, as before, please follow the PEP 257 rules 🙏🏽 . Pay attention : The summary line must fit in one line (88chars)

histolab/tile.py
```diff
@@ -168,7 +168,22 @@ def save(self, path: Union[str, bytes, os.PathLike]) -> None:
...
+        np.ndarray
+            Binary mask representing the tissue in the tile.
+        """
+        return self.calculate_tissue_mask(FiltersComposition(Tile))
+
+    def calculate_tissue_mask(
+        self, filter_composition: FiltersComposition
+    ) -> np.ndarray:
+        """Calculate the binary mask representing the tissue in the tile
+        given FiltersComposition.
```
ernestoarbitrio, line 186: same as before for the PEP.

ernestoarbitrio, line 186:
```suggestion
        """Calculate the binary mask representing the tissue in the tile
        given ``filter_composition``.
```

tests/unit/test_compositions.py
```diff
@@ -62,6 +66,24 @@ def it_knows_tissue_areas_mask_slide_filters_composition(
...
+):
+    _enough_tissue_mask_filters_ = FiltersComposition(
+        Compose,
+        RgbToGrayscale_,
+        OtsuThreshold_,
+        BinaryDilation_,
+        RemoveSmallHoles_,
+        RemoveSmallObjects_,
+    ).tissue_mask_filters
+    assert type(_enough_tissue_mask_filters_) == Compose
```
ernestoarbitrio, line 84: IMHO this test is not meaningful. If you take a look at lines 52-58 you will see the assertions for the mocks. Could you please integrate those also in this test? Thanks

tests/unit/test_masks.py
```diff
@@ -1,7 +1,11 @@
 import numpy as np
 import pytest
 
-from histolab.filters.compositions import _SlideFiltersComposition
+from histolab.filters import image_filters
```
ernestoarbitrio, line 4: please could avoid importing the whole module? I'd like to see the import statements only for the object needed 😄

tests/unit/test_masks.py
```diff
@@ -138,6 +142,24 @@ def it_knows_its_mask_slide(
 
         assert binary_mask == expected_mask
 
+    def it_knows_its_mask_slide_custom(self, request, tmpdir):
+        slide, _ = base_test_slide(tmpdir, PILIMG.RGBA_COLOR_500X500_155_249_240)
+        custom_filters = [image_filters.RgbToGrayscale()]
+
```
ernestoarbitrio, line 148:
```suggestion
```

tests/unit/test_masks.py
```diff
@@ -153,6 +175,39 @@ def it_knows_its_mask_tile(self, request):
 
         assert binary_mask == expected_mask
 
+    def it_knows_its_mask_tile_and_compute_with_default(self, request):
+        tile = Tile(PILIMG.RGBA_COLOR_500X500_155_249_240, None, None)
+
```
ernestoarbitrio, line 180:
```suggestion
```

tests/unit/test_masks.py
```diff
@@ -153,6 +175,39 @@ def it_knows_its_mask_tile(self, request):
...
+        tissue_mask = TissueMask()
+        _ = tissue_mask(tile)
+
+        calculate_tissue_mask_tile_.assert_called_once()
+
+    def it_knows_its_mask_tile_and_supports_custom_filters(self, request):
+        tile = Tile(PILIMG.RGBA_COLOR_500X500_155_249_240, None, None)
+        custom_filter_cls = image_filters.RgbToGrayscale
+
```
ernestoarbitrio, line 193:
```suggestion
```

tests/unit/test_masks.py
```diff
@@ -153,6 +175,39 @@ def it_knows_its_mask_tile(self, request):
...
+        custom_filters = [custom_filter_cls()]
+        custom_filter_call_ = method_mock(request, custom_filter_cls, "__call__")
+
```
ernestoarbitrio, line 196:
```suggestion
```

tests/unit/test_masks.py
```diff
@@ -138,6 +142,24 @@ def it_knows_its_mask_slide(
...
+            request, _CustomFiltersComposition, "set_custom_filters"
+        )
+        custom_filter_call_ = method_mock(
+            request, image_filters.RgbToGrayscale, "__call__"
+        )
+
+        tissue_mask = TissueMask(*custom_filters)
+        _ = tissue_mask(slide)
+
+        set_custom_filters_.assert_called_once()
```
ernestoarbitrio, line 159: I'm not a huge fan of `assert_called_once`. The reason is that, yes, you're asserting that a proper method has been called, but the interesting info would be also know with which parameters? I'd suggest to switch to `assert_called_once_with` and check the calling args 😉

tests/unit/test_masks.py
```diff
@@ -138,6 +142,24 @@ def it_knows_its_mask_slide(
...
+        assert set_custom_filters_.call_args[0][1] == custom_filters[0]
+        custom_filter_call_.assert_called_once()
```
ernestoarbitrio, line 161: same

tests/unit/test_masks.py
```diff
@@ -153,6 +175,39 @@ def it_knows_its_mask_tile(self, request):
...
+        tile = Tile(PILIMG.RGBA_COLOR_500X500_155_249_240, None, None)
+
+        calculate_tissue_mask_tile_ = method_mock(
+            request, Tile, "calculate_tissue_mask"
+        )
+
+        tissue_mask = TissueMask()
+        _ = tissue_mask(tile)
+
+        calculate_tissue_mask_tile_.assert_called_once()
```
ernestoarbitrio, line 188: same

tests/unit/test_masks.py
```diff
@@ -153,6 +175,39 @@ def it_knows_its_mask_tile(self, request):
...
+            request, _CustomFiltersComposition, "set_custom_filters"
+        )
+
+        # Call the tissue mask with custom filters
+        tissue_mask = TissueMask(*custom_filters)
+        _ = tissue_mask(tile)
+
+        # Ensure the custom tissue calculator, the internal setter
+        # and the custom filter itself is called
+        set_custom_filters_.assert_called_once()
```
ernestoarbitrio, line 207: same

tests/unit/test_masks.py
```diff
@@ -153,6 +175,39 @@ def it_knows_its_mask_tile(self, request):
...
+        assert set_custom_filters_.call_args[0][1] == custom_filters[0]
+        custom_filter_call_.assert_called_once()
```
ernestoarbitrio, line 209: same

ernestoarbitrio (changes requested):
@Christopher22 I did this CR round with @alessiamarcolini.
Sorry for the bunch of request changes ... but overall looks neat.
Could you please address the comments in order to make us able to do a final round?
Thanks again for you effort 💪🏽 ... we really appreciated it 😄

Christopher22:
Thank you so much for your comments! I hope, I was able to fix the things up you mentioned.

As a small refactor, I removed my 'set_custom_filters' and replaced it with a meaningful constructor which fits better in your existing codebase.

tests/unit/test_masks.py
Christopher22: I try to remove all of the things. However, I sometimes have no access to the "self" parameter. In this case, I try to check only the remaining arguments.

ernestoarbitrio: Sorry I do not understand this point. Could you elaborate? Thx

histolab/masks.py
```diff
@@ -178,8 +191,14 @@ def _mask(self, slide) -> np.ndarray:
             Binary mask of the tissue area. The dimensions are those of the thumbnail.
         """
         thumb = slide.thumbnail
-        filters = FiltersComposition(histolab.slide.Slide).tissue_mask_filters
-        thumb_mask = filters(thumb)
+
+        # Generate appropriate composition of filter
+        if len(self.custom_filters) == 0:
+            composition = FiltersComposition(histolab.slide.Slide)
```
ernestoarbitrio, line 197: Could you import directly `Slide` and use it instead of the whole path?

tests/unit/test_masks.py
```diff
@@ -148,11 +159,39 @@ def it_knows_its_mask_tile(self, request):
...
 
         tissue_mask = TissueMask()
-
         binary_mask = tissue_mask(tile)
 
         assert binary_mask == expected_mask
 
+    def it_knows_its_mask_tile_and_compute_with_default(self, request):
+        tile = Tile(PILIMG.RGBA_COLOR_500X500_155_249_240, None, None)
+
```
ernestoarbitrio, line 168: i would remove this line. The reason is that in a test, for the sake of readability and for future developers we're trying to keep separated the 3 blocks that compose a test.
1. Objects declarations
2. Actual call
3. Assertions

Hope this can help

ernestoarbitrio (changes requested):
Hey @Christopher22 thanks for all your commitment. I left a couple of furhter minor comments. Could you please take a look? After that we are probably ready to go.

Heads up to @alessiamarcolini ☝🏽

Christopher22:
Amazing! Your highlighed areas should be fixed by now :)

histolab/filters/compositions.py
```diff
@@ -29,7 +31,7 @@ class FiltersComposition:
 
     Arguments
     ---------
-    cls_ : type, {Tile, Slide}
+    cls_ : type, {Tile, Slide, Compose}
```
alessiamarcolini, line 34: docs: `*custom_filters parameter` is missing

histolab/filters/compositions.py
```diff
@@ -41,12 +43,13 @@ class FiltersComposition:
         >>> filters_tile = FiltersComposition(Tile).tissue_mask_filters
     """
 
-    def __new__(cls: type, cls_):
+    def __new__(cls: type, cls_, *custom_filters: imf.ImageFilter):
```
alessiamarcolini, line 46:
```suggestion
    def __new__(cls, cls_,: type *custom_filters: imf.ImageFilter):
```

histolab/filters/compositions.py
```diff
@@ -139,3 +142,35 @@ def tissue_mask_filters(self) -> imf.Compose:
...
+        self, cls_: Type[imf.Compose], *custom_filters: imf.ImageFilter
+    ) -> None:
+        """Create a custom filter composition with at least one filter.
+
+        Generally, their is no need to call this constructor manually.
+        Please use ``FiltersComposition(Compose, *filters)`` instead.
+        """
+
+        if cls_ != imf.Compose:
+            raise FilterCompositionError("cls_ parameter cannot be None")
```
alessiamarcolini, line 158: as Codecov says, this line is not covered by tests (in this case it should be a unit test)

histolab/filters/compositions.py
```diff
@@ -139,3 +142,35 @@ def tissue_mask_filters(self) -> imf.Compose:
...
+
+        if len(custom_filters) == 0:
+            raise FilterCompositionError("'Compose' requires at least one filter")
```
alessiamarcolini, line 161: same here, not covered by tests

histolab/filters/compositions.py
```diff
@@ -139,3 +142,35 @@ def tissue_mask_filters(self) -> imf.Compose:
...
+    @lazyproperty
+    def tissue_mask_filters(self) -> imf.Compose:
+        """Filters composition for tissue estimation with custom filters.
+
+        Returns
+        -------
+        imf.Compose
+            Filters composition
+        """
+        if len(self.custom_filters) == 0:
```
alessiamarcolini, line 174: Is this possible? If yes, it needs a test

histolab/masks.py
```diff
@@ -132,10 +133,22 @@ def _regions(regions: List[Region], n: int = 1) -> List[Region]:
...
+        """
+        Create a new tissue mask.
+
+        If custom image filters are specified, those are used instead the default ones.
+        By default, the tissue within the slide or tile is automatically detected
+        through a predefined chain of filters.
+
+        Parameters
+        ----------
+        filters : ImageFilter
```
alessiamarcolini, line 148:
```suggestion
        *filters : ImageFilter
```
Otherwise I wouldn't know how to specify that that's an arbitrary number of parameters

tests/unit/test_masks.py
```diff
@@ -148,11 +159,38 @@ def it_knows_its_mask_tile(self, request):
         tissue_mask_tile_.return_value = expected_mask
```
alessiamarcolini, line 160: not your responsibility, but since you touched this method... this line should also be removed

tests/unit/test_masks.py
```diff
@@ -148,11 +159,38 @@ def it_knows_its_mask_tile(self, request):
...
+    def it_knows_its_mask_tile_and_compute_with_default(self, request):
+        tile = Tile(PILIMG.RGBA_COLOR_500X500_155_249_240, None, None)
+        calculate_tissue_mask_tile_ = method_mock(
+            request, Tile, "calculate_tissue_mask"
+        )
+
+        tissue_mask = TissueMask()
+        _ = tissue_mask(tile)
+
+        calculate_tissue_mask_tile_.assert_called_once()
```
alessiamarcolini, line 175: the problem of not checking the parameters can hide bugs in this very case: this test is not really checking that the mask is computed with the default filters. The method `Tile.calculate_tissue_mask` could be called:
1. by `Tile.tissue_mask`, with `FiltersComposition(Tile)` as parameter, if the code logic is correct OR
2. by _mask (dispatched to the `Tile`'s method `_`) with `custom_filters`, if the logic is not correct.

So please add the parameters in the assertion :)

tests/unit/test_masks.py
alessiamarcolini: Regarding the whole test: I feel that testing that `RgbToGrayscale` is called is not a good strategy, because:

1. usually we want to mock all the methods/instances that get called/used directly by the method we are testing, so in this case it would be `Tile.calculate_tissue_mask`
2. as a proof for my first point: this test is not really checking that the mask is built with the custom filters, because `RgbToGrayscale` is actually the first filter in the "default" filter sequence (for both `Slide` and Tile). So this test would actually pass with both the default and the "custom" filters represented by `[RgbToGrayscale()]` 🤓

Please let me know if this makes sense for you!

alessiamarcolini (changes requested):
Apart from a couple of docstrings improvements, I had to reason carefully about two unit tests (see comments) and please try to cover all of your new code with tests (follow codecov comments), as we cannot merge this PR if it does not provide 100% coverage ✌🏼

Christopher22:
Such cute apes! ;) Your comments should be integrated by now!

tests/unit/test_masks.py
alessiamarcolini: @Christopher22 I think you misunderstood my points here. This updated test covers point number 2 but not number 1, the mocked object is not directly called/used by the method you're testing.

alessiamarcolini (changes requested):
Please check the comment I left about the `it_knows_its_mask_tile_and_supports_custom_filters` test.

PS please squash the commits at the end of the review, otherwise it's more difficult for us to be on track with the changes between review rounds ✌🏼

Christopher22:
@alessiamarcolini Do you prefer a test such as this instead? :)

ernestoarbitrio (approved):

tests/unit/test_masks.py
```diff
@@ -177,22 +180,30 @@ def it_knows_its_mask_tile(self, request):
...
         tile = Tile(PILIMG.RGBA_COLOR_500X500_155_249_240, None, None)
         custom_filters = [CustomFilterForTest()]
-        custom_filter_call_ = method_mock(request, CustomFilterForTest, "__call__")
-        # We need to account for the fact padding is added before filtering
-        expected_mask = np.zeros((510, 510), dtype=np.bool)
-        custom_filter_call_.return_value = expected_mask
+        expected_mask = [
+            [True, True],
+            [False, True],
+        ]
```
alessiamarcolini, line 186: please use a np array here to be consistent with the real `calculate_tissue_mask` return type

tests/unit/test_masks.py
```diff
@@ -177,22 +180,30 @@ def it_knows_its_mask_tile(self, request):
...
+        assert len(calculate_tissue_mask_call.call_args[0]) == 2
+        assert calculate_tissue_mask_call.call_args[0][0] == tile
+        assert isinstance(
+            calculate_tissue_mask_call.call_args[0][1], _CustomFiltersComposition
+        )
+        assert calculate_tissue_mask_call.call_args[0][1]._custom_filters == tuple(
+            custom_filters
+        )
+
+        assert binary_mask == expected_mask
```
alessiamarcolini, line 206: accordigly, here please use `np.testing.assert_array_equal`

alessiamarcolini (changes requested):
Just a minor comment for consistency, the rest is perfect!! 👏🏼

Christopher22:
Ready for squash ;)

alessiamarcolini (approved):
Let's merge it!

alessiamarcolini:
@Christopher22 thank you so much for your valuable contribution! 👏🏼👏🏼

Now, to have a consistent code base, we would like to provide the same flexibility for the `BiggestTissueBoxMask` as well. Please let us know if you are interested in contributing again in this regard! Otherwise, we can take care of it, no pressure at all 😉

PR #380, opened 2022-01-28
Fix type hints

patrick91:
This PR attemps to fix some of (or all) the typing issues found by running `mypy histolab`.

Unfortunately some of these issues are hard to fix without context and without making a design decision. I've left a couple of TODOs in the code for that.

The biggest issue is having `class Filter(Protocol)` that has a call method that accepts either an image or a ndarray, unfortunaly mypy doesn't allow to change the argument type to a more specific one (see https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides). Maybe we should have two types of Filters, one for images and one for ndarrays.

Also, I'm not 100% convinced Protocol is useful here (and in the Tiler class), what's the reasoning behind using Protocol and not abc.ABC?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -144,6 +144,11 @@ jobs:
         run: |
           flake8 . --config=setup.cfg --count --statistics
 
+      - name: mypy
+        id: black
+        run: |
+          mypy histolab .
+
       - name: Check Vulnerabilities with Bandit
         id: bandit
         run: |
--- a/histolab/filters/image_filters.py
+++ b/histolab/filters/image_filters.py
@@ -19,32 +19,31 @@
 object. Additionally, some of the image filters in histolab leverage functions and
 utilities by scikit-image. Image filters are divided into sub-categories, depending on
 their behaviour and output type.
- """
+"""
 
 import operator
 from abc import abstractmethod
 from typing import Any, Callable, List, Union
 
 import numpy as np
-import PIL
+from PIL import  ImageOps
+import PIL.Image
+from typing_extensions import Protocol, runtime_checkable
 
 from .. import util
 from . import image_filters_functional as F
 
-try:
-    from typing import Protocol, runtime_checkable
-except ImportError:
-    from typing_extensions import Protocol, runtime_checkable
+ImageOrNPArray = Union[PIL.Image.Image, np.ndarray]
 
 
+# TODO: should we remove this and only keep ImageFilter?
+# TODO: does it make sense to pass an np.ndarray as input?
 @runtime_checkable
 class Filter(Protocol):
     """Filter protocol"""
 
     @abstractmethod
-    def __call__(
-        self, img: Union[PIL.Image.Image, np.ndarray]
-    ) -> Union[PIL.Image.Image, np.ndarray]:
+    def __call__(self, img: ImageOrNPArray) -> ImageOrNPArray:
         pass  # pragma: no cover
 
     def __repr__(self) -> str:
@@ -56,7 +55,7 @@ class ImageFilter(Filter, Protocol):
     """Image filter protocol"""
 
     @abstractmethod
-    def __call__(self, img: PIL.Image.Image) -> Union[PIL.Image.Image, np.ndarray]:
+    def __call__(self, img: ImageOrNPArray) -> ImageOrNPArray:
         pass  # pragma: no cover
 
 
@@ -72,7 +71,7 @@ class Compose(ImageFilter):
     def __init__(self, filters: List[Filter]) -> None:
         self.filters = filters
 
-    def __call__(self, img: PIL.Image.Image) -> Union[PIL.Image.Image, np.ndarray]:
+    def __call__(self, img: ImageOrNPArray) -> ImageOrNPArray:
         for filter_ in self.filters:
             img = filter_(img)
         return img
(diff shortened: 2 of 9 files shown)


.github/workflows/tests.yml
```diff
@@ -144,6 +144,11 @@ jobs:
         run: |
           flake8 . --config=setup.cfg --count --statistics
 
+      - name: mypy
+        id: black
+        run: |
+          mypy histolab .
```
patrick91, line 150: we should move all these steps outside the matrix, they should run only once 😊

histolab/filters/morphological_filters.py
```diff
@@ -24,10 +24,8 @@
 from . import morphological_filters_functional as F
 from .image_filters import Filter
 
-try:
-    from typing import Protocol, runtime_checkable
-except ImportError:
-    from typing_extensions import Protocol, runtime_checkable
+
+from typing_extensions import Protocol, runtime_checkable
```
alessiamarcolini, line 28: I think that we were using this try/except because we are still supporting python 3.7, and Protocol has been introduced in Python 3.8

histolab/filters/image_filters.py
```diff
@@ -99,7 +98,9 @@ def __init__(self, lambd: Callable[[PIL.Image.Image], PIL.Image.Image]) -> None:
         assert callable(lambd), repr(type(lambd).__name__) + " object is not callable"
         self.lambd = lambd
 
-    def __call__(self, img: PIL.Image.Image) -> Union[PIL.Image.Image, np.ndarray]:
+    def __call__(self, img: ImageOrNPArray) -> ImageOrNPArray:
+        assert isinstance(img, PIL.Image.Image), "Input must be an image"
```
alessiamarcolini, line 102: best practice: we don't want to have assertion statements in code but only in the tests, better to raise an appropriate exception

histolab/filters/image_filters.py
```diff
@@ -117,7 +118,9 @@ class ToPILImage(Filter):
         The image represented as PIL Image
     """
 
-    def __call__(self, np_img: np.ndarray) -> PIL.Image.Image:
+    def __call__(self, np_img: ImageOrNPArray) -> PIL.Image.Image:
+        assert isinstance(np_img, np.ndarray), "Input must be a numpy array"
```
alessiamarcolini, line 122: same here

histolab/filters/image_filters.py
```diff
@@ -137,10 +140,12 @@ class ApplyMaskImage(Filter):
         Image with the mask applied
     """
 
-    def __init__(self, img: PIL.Image.Image) -> None:
+    def __init__(self, img: ImageOrNPArray) -> None:
+        assert isinstance(img, PIL.Image.Image), "Input must be an image"
```
alessiamarcolini, line 144: same here

histolab/filters/image_filters.py
```diff
@@ -137,10 +140,12 @@ class ApplyMaskImage(Filter):
...
         self.img = img
 
-    def __call__(self, mask: np.ndarray) -> PIL.Image.Image:
+    def __call__(self, mask: ImageOrNPArray) -> PIL.Image.Image:
+        assert isinstance(mask, np.ndarray), "Mask must be a numpy array"
```
alessiamarcolini, line 148: same here

histolab/filters/image_filters.py
```diff
@@ -178,7 +183,8 @@ class Invert(ImageFilter):
         >>> image_inv_gray = invert(image_gray)
     """  # noqa
 
-    def __call__(self, img: PIL.Image.Image) -> PIL.Image.Image:
+    def __call__(self, img: ImageOrNPArray) -> PIL.Image.Image:
+        assert isinstance(img, PIL.Image.Image), "Input must be an image"
```
alessiamarcolini, line 187: well, for all the asserts in the module :D

histolab/util.py
```diff
@@ -351,6 +355,6 @@ def method_dispatch(func: Callable[..., Any]) -> Callable[..., Any]:
     def wrapper(*args, **kw):
         return dispatcher.dispatch(args[1].__class__)(*args, **kw)
 
-    wrapper.register = dispatcher.register
+    setattr(wrapper, "register", dispatcher.register)
```
alessiamarcolini, line 358: curiosity: why this?

alessiamarcolini (changes requested):
Hey @patrick91 thank you very much for this contribution! ❤️
I just gave a quick look and left some comments. I will need to get back to this :)

We definitely need to rethink some of the incongruences - thanks for leaving the TODOs in the code

.github/workflows/tests.yml
alessiamarcolini: good point

histolab/filters/morphological_filters.py
patrick91: yes, that's why I'm importing it directly from typing exceptions, it is a dependency of histolab os it is fine to just do that.

Doing the try catch will complicate the typing quite a bit unfortunately

histolab/filters/image_filters.py
patrick91: oh the assert here is just for making mypy happy :D I think it will go away with some refactoring 😊

histolab/filters/morphological_filters.py
alessiamarcolini: fair enough

histolab/util.py
patrick91: doing `wrapper.register = xxx` throws a mypy error because `register` doesn't exist on wrapper so as a workaround we use setattr 😊

PR #466, opened 2022-11-13
update scikit-image to 0.19.3

ernestoarbitrio:
with this input image
diagnostic-slide-thumb-gs

the old expecation is 
old

the new one
new

I do believe the new one with 0.19.3 is better cause it doesn't have the withe tiny border, but I need your feedback @alessiamarcolini @nicolebussola


nicolebussola (approved):
LGTM

alessiamarcolini (approved):

PR #468, opened 2022-11-19
Dependencies update

ernestoarbitrio:
**UPDATE LARGE IMAGE**

W/O large image
false
WITH
true

The behaviour is the same as before. The only thing I had to do is to preserve the aspect ratio for thumb size when `use_largeimage` is true cause in the last version introduced some changes.

I do believe my changes are legit and don't introduce any regression.


tests/unit/test_slide.py
```diff
@@ -175,7 +175,7 @@ def it_has_largeimage_tilesource(self, tmpdir):
             tmpdir, PILIMG.RGBA_COLOR_500X500_155_249_240, use_largeimage=True
         )
 
-        assert slide._tile_source.name == "pilfile"
+        assert slide._tile_source.name == "pil"
```
ernestoarbitrio, line 178: legit change due to this commit 
https://github.com/girder/large_image/commit/8f7e889e56cba6f6339d4e85659921e0626612c2

PR #494, opened 2023-01-14
Bump large image

ernestoarbitrio:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

# DETAILS
**UPDATE LARGE IMAGE**

W/O large image
false
WITH
true

The behaviour is the same as before. The only thing I had to do is to preserve the aspect ratio for thumb size when `use_largeimage` is true cause in the last version introduced some changes.

I do believe my changes are legit and don't introduce any regression.


histolab/util.py
```diff
@@ -90,6 +91,40 @@ def _transform_float(img_array: np.ndarray) -> np.ndarray:
...
 
 
+def refine_thumbnail_size_preserving_aspect_ratio(
+    thumbnail_size: Tuple[int, int], original_size: Tuple[int, int]
+) -> Tuple[int, int]:
+    """Return thumbnail dimensions preserving the original image aspect ratio.
+
+    Parameters
+    ----------
+    thumbnail_size: tuple
```
alessiamarcolini, line 101:
```suggestion
    thumbnail_size : Tuple[int, int]
```

histolab/util.py
```diff
@@ -90,6 +91,40 @@ def _transform_float(img_array: np.ndarray) -> np.ndarray:
...
+       The width and height of the thumbnail.
+    original_size: tuple
```
alessiamarcolini, line 103:
```suggestion
    original_size : Tuple[int, int]
```

histolab/util.py
```diff
@@ -90,6 +91,40 @@ def _transform_float(img_array: np.ndarray) -> np.ndarray:
...
+       The width and height of the original image.
+
+    Returns
+    -------
+       The news width and height with aspect ration preserved.
```
alessiamarcolini, line 108:
```suggestion
       The new width and height with preserved aspect ratio.
```

alessiamarcolini (approved):

PR #539, opened 2023-03-21
Added ThresholdTiler

erich-r:
## Description
Added a new tiler that, given a scorer and a threshold, extracts all tiles whose score is above the threshold.
There may be instances when the user does not want the top _n_ tiles from a slide, but rather every tile with a score over a specified threshold.
In this instance, the user does not know beforehand how many tiles must be extracted. 

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

## Checklist


histolab/tiler.py
```diff
@@ -1121,3 +1121,288 @@ def _scores(
...
+        all_scores = self._scores(slide, extraction_mask)
+        scaled_scores = self._scale_scores(all_scores)
+
+        if self.threshold < 0:
+            raise ValueError(f"'threshold' cannot be negative ({self.threshold})")
+
+        filtered_tiles_scores = []
+        filtered_tiles_scaled_scores = []
+
+        for i in range(len(all_scores)):
```
erich-r, line 1304: There is no need to sort the elements by their score, it is sufficient to iterate the elements and save only the elements whose score is above the threshold

tests/unit/test_tiler.py
```diff
@@ -1253,3 +1253,376 @@ def it_can_extract_score_tiles_and_save_report(self, request, tmpdir):
...
+        assert result == expected_result
+
+    def it_can_calculate_scores(self, request):
+        slide = instance_mock(request, Slide)
+        coords = CP(0, 0, 10, 10)
+        image = PILIMG.RGB_RANDOM_COLOR_500X500
+        tile = Tile(image, coords)
+        _tiles_generator = method_mock(request, GridTiler, "_tiles_generator")
+        # it needs to be a generator
+        _tiles_generator.return_value = ((tile, coords) for i in range(3))
```
ernestoarbitrio, line 1311: The `_tiles_generator` return value in the signatures is `Tuple[List[Tuple[float, CoordinatePair]], List[Tuple[float, CoordinatePair]]]` and here you are assigning `(tile, coords)` I don't think it's quite correct.
So or the test is wrong or the signature is wrong. :D

histolab/tiler.py
```diff
@@ -1121,3 +1121,288 @@ def _scores(
...
+
+        if report_path:
+            self._save_report(
+                report_path,
+                filtered_tiles_scores,
+                filtered_tiles_scaled_scores,
+                filenames,
+            )
+
+        logger.info(f"{tiles_counter+1} Grid Tiles have been saved.")
```
ernestoarbitrio, line 1258: why the +1?

histolab/tiler.py
```diff
@@ -1121,3 +1121,288 @@ def _scores(
...
+        tiles_counter = 0
+        filenames = []
+
+        for tiles_counter, (score, tile_wsi_coords) in enumerate(filtered_tiles_scores):
+            tile = slide.extract_tile(
+                tile_wsi_coords,
+                tile_size=self.final_tile_size,
+                mpp=self.mpp,
+                level=self.level if self.mpp is None else None,
+            )
```
ernestoarbitrio, line 1242: IMHO the tile should be extracted in the `_tiles_generator` and then iterating over the tiles as we do in the `GridTiler`.

histolab/tiler.py
```diff
@@ -1121,3 +1121,288 @@ def _scores(
             scores.append((score, tile_wsi_coords))
 
         return scores
+
+
+class ThresholdTiler(GridTiler):
```
ernestoarbitrio, line 1126: Sorry, but I d not understand why this is a subcalss of the `GridTiler`. `extract` and `_tiles_generator` have been reimplemented and no other methods or properties from the `GridTiler` have been used. Furthermore you have also changed the init method, so personally I find weird the `GridTiler` dependency. At a first glance it seems more similar to the `ScoreTiler`.

ernestoarbitrio (changes requested):
Hi @erich-r 👋

Thanks for your contribution, we really appreciate it. ❤️

I had a few time to do a first round of code review and left some comments here.

Please take a look at my comments and also I believe that you miss the **integration tests** for this new tiler. **Unit tests are not enough to prevent regressions.**

My comments are mostly implementation and design wise ... waiting for @alessiamarcolini or @nicolebussola for a paired CR.

histolab/tiler.py
erich-r: > why the +1?

Because tiles_counter starts at 0.
If you want to print how many iterations you have done you need to add 1, or am I missing something?

ernestoarbitrio: what if tiles_counters remains 0 ? It's an edge case but still possible IMHO

tests/unit/test_tiler.py
erich-r: The method being mocked is `_tiles_generator` of the `GridTiler` class, which I think has the signature `Tuple[Tile, CoordinatePair]`, so I think both are correct (?)

erich-r: > Oh ok, but now the question is why are you mocking the `GridTiler`?

Because the aim of the test is to check whether the tiler can extract the scores from tiles, so I think it does not matter if the tiles come from a `GridTiler`.

In the test `it_can_calculate_filtered_tiles` I am calling the `_tiles_generator` from a `ThresholdTiler`.

histolab/tiler.py
erich-r: > Sorry, but I d not understand why this is a subcalss of the `GridTiler`. `extract` and `_tiles_generator` have been reimplemented and no other methods or properties from the `GridTiler` have been used.

The reason is a subclass of GridTiler is because ThresholdTiler first extracts tiles like a GridTiler (calling its method `_tiles_generator` line 1402 of `tiler.py`), then for each tile it assigns a related score (line 1406), and finally all tiles that have a score less than a specified threshold are discarded.

> At a first glance it seems more similar to the `ScoreTiler`.

Did you mean inheriting from `ScoreTiler`?
If yes, the problem with inheriting from `ScoreTiler` is that, when instantiating a `ThresholdTiler`, the user must supply `n_tiles` as a parameter, which I believe is questionable for a `ThresholdTiler` since the user does not know in advance how many tiles must be removed (he's more interested in all tiles that have a score higher than a threshold).

ernestoarbitrio: gotcha

nicolebussola:
Hi Erich, thank you for your contribution!  I think it could be more intuitive to select a threshold when the scores are scaled between 0 and 1 rather than on raw scores, what do you think?

nicolebussola:
Hi Erich, thank you for your contribution!  I think it could be more intuitive to select a threshold when the scores are scaled between 0 and 1 rather than on raw scores, what do you think?

histolab/tiler.py
```diff
@@ -1121,3 +1121,292 @@ def _scores(
...
+        if self.threshold < 0:
+            raise ValueError(f"'threshold' cannot be negative ({self.threshold})")
+
+        filtered_tiles_scores = []
+        filtered_tiles_scaled_scores = []
+
+        for i in range(len(all_scores)):
+            if all_scores[i][0] > self.threshold:
+                filtered_tiles_scores.append(all_scores[i])
+                filtered_tiles_scaled_scores.append(scaled_scores[i])
```
alessiamarcolini, line 1311: This could be refactored into this, to make it more Pythonic

```suggestion
        for score in all_scores:
            if score[0] > self.threshold:
                filtered_tiles_scores.append(score)
                filtered_tiles_scaled_scores.append(score)
```

histolab/tiler.py
```diff
@@ -1121,3 +1121,292 @@ def _scores(
...
+    @staticmethod
+    def _save_report(
+        report_path: str,
+        filtered_tiles_scores: List[Tuple[float, CoordinatePair]],
+        filtered_tiles_scaled_scores: List[Tuple[float, CoordinatePair]],
+        filenames: List[str],
+    ) -> None:
+        """Save to ``filename`` the report of the saved tiles with the associated score.
+
+        The CSV file
```
alessiamarcolini, line 1324: oh, I see this line is cut (same as in the `ScoreTiler`). I'm not sure what we wanted to write here (?)
Could you remove this from both here and the `ScoreTiler`?

tests/integration/test_tiler.py
```diff
@@ -515,3 +515,89 @@ def it_locates_tiles_on_the_slide(
...
+        request,
+        fixture_slide,
+        threshold,
+        tile_size,
+        level,
+        check_tissue,
+        expectation,
+    ):
+        slide = Slide(fixture_slide, "")
+        scored_tiles_extractor = ThresholdTiler(
```
alessiamarcolini, line 577: maybe we want to call it `thresholded_tiles_extractor` to be more on the theme?

tests/unit/test_tiler.py
alessiamarcolini: I agree with @ernestoarbitrio: here we should mock the `_tiles_generator` from the `ThresholdTiler` to be more precise (and the tests for the `ScoreTiler` should do the same, oops) and assign the correct return value

tests/unit/test_tiler.py
```diff
@@ -1253,3 +1253,376 @@ def it_can_extract_score_tiles_and_save_report(self, request, tmpdir):
...
+        assert type(scores[0]) == tuple
+        assert type(scores[0][0]) == float
+        assert type(scores[0][1]) == CP
+        assert scores == [(0.5, coords), (0.7, coords)]
+
+    def but_it_raises_runtimeerror_if_no_tiles_are_extracted(self, request):
+        slide = instance_mock(request, Slide)
+        _tiles_generator = method_mock(request, GridTiler, "_tiles_generator")
+        # it needs to be an empty generator
+        _tiles_generator.return_value = (n for n in [])
```
alessiamarcolini, line 1334: same here

tests/unit/test_tiler.py
```diff
@@ -1253,3 +1253,376 @@ def it_can_extract_score_tiles_and_save_report(self, request, tmpdir):
...
+        binary_mask = BiggestTissueBoxMask()
+
+        with pytest.raises(ValueError) as err:
+            threshold_tiler.extract(slide, binary_mask)
+
+        assert isinstance(err.value, ValueError)
+        assert str(err.value) == "'threshold' cannot be negative (-1)"
+        _scores.assert_called_once_with(threshold_tiler, slide, binary_mask)
+
+    def it_can_extract_score_tiles(self, request, tmpdir):
```
alessiamarcolini, line 1466:
```suggestion
    def it_can_extract_thresholded_tiles(self, request, tmpdir):
```

tests/unit/test_tiler.py
```diff
@@ -1253,3 +1253,376 @@ def it_can_extract_score_tiles_and_save_report(self, request, tmpdir):
...
+            ThresholdTiler,
+            "_tiles_generator",
+        )
+        tile = Tile(image, coords)
+        _extract_tile.return_value = tile
+        _tiles_generator.return_value = (
+            [(0.8, coords), (0.7, coords)],
+            [(0.8, coords), (0.7, coords)],
+        )
+        _tile_filename = method_mock(request, GridTiler, "_tile_filename")
```
alessiamarcolini, line 1485:
```suggestion
        _tile_filename = method_mock(request, ThresholdTiler, "_tile_filename")
```

alessiamarcolini (changes requested):
Hi @erich-r thank you so much for your contribution ⭐ 

As @ernestoarbitrio pointed out, it's a bit weird that the `ThresholdTiler` inherits from the `GridTiler` and not the `ScoreTiler` - though I understand that the fact that the `ScoreTiler` imposes a `n_tiles` parameter and that is not acceptable in this case. 

Anyway this is bringing a lot of code duplication, which makes me wonder whether we should move the common code to some other "support" Tiler, or move out the static methods to standalone functions... What do we think?

ernestoarbitrio:
Hi 👋,
   just checking @erich-r do you have any news on this PR?

erich-r:
> Hi wave, just checking @erich-r do you have any news on this PR?

Hi! 
- I wrote the integration tests
- Now the threshold is being compared with the normalized scores (as suggested by @nicolebussola )
Next I'm going to apply @alessiamarcolini suggestions (as soon as possible)

ernestoarbitrio:
> > Hi wave, just checking @erich-r do you have any news on this PR?
> 
> Hi!
> 
> * I wrote the integration tests
> * Now the threshold is being compared with the normalized scores (as suggested by @nicolebussola )
>   Next I'm going to apply @alessiamarcolini suggestions (as soon as possible)

awesome thanks a lot

alessiamarcolini:
hey @erich-r do you have any updates about this PR?

PR #709, opened 2024-10-01
fix ci

ernestoarbitrio:


.github/workflows/tests.yml
```diff
@@ -50,25 +50,44 @@ jobs:
...
+            python-version: "3.11"
+            cache-path: ~/Library/Caches/pypoetry/cache
+          - os: macOS-13
+            python-version: "3.8"
+            cache-path: ~/Library/Caches/pypoetry/cache
+          - os: macOS-13
+            python-version: "3.9"
+            cache-path: ~/Library/Caches/pypoetry/cache
+          - os: macOS-13
+            python-version: "3.10"
```
alessiamarcolini, line 77: it's super hard to understand what's in and what's out... can we figure out a better way to deal with this?

ernestoarbitrio: we can but ATM I'm just trying to fix it :D when I have a better code I'll ask you the CR 💋

ernestoarbitrio: macOs-latest runners are now on arm64 and I'm still having issues on make openslide working with this env even if on my macbook works fine 🤷🏽‍♂️ . ATM all the macOS-runners are on the latest x86 that is macos-13. IMHO we can keep it as is until we discover what's the issue with arm and github runners.

alessiamarcolini: ok fine for now... maybe we can document this behavior by adding macOS arm to the CI but setting it as experimental so it can fail without failing the CI

alessiamarcolini (approved):

PR #833, opened 2026-04-18
Remove openslide in favor of openslide-bin and support macos-15

alessiamarcolini:
## Description
Per title. With `openslide-bin`, we don't need to manage the pixman version ourselves, as it ships with v0.46.4-1, and we don't risk accidentally loading v0.38 when provided by the OS

With macos-15 running on arm + changes in low-level C libraries bundled in `openslide-bin`, we notice some off-by-one errors in the tests where we render a thumbnail of the slide then draw rectangles/overlays indicating masks or tile positions.
The failures were due to ultra‑strict, pixel‑perfect PNG comparisons detecting insignificant 1‑intensity differences in a vanishingly small fraction of pixels, almost certainly caused by platform‑specific rounding/compositing in Pillow/NumPy rather than any OpenSlide/pixman bug.
So we relax those tests to allow these rounding errors.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

* Fixes:

## Checklist

- [ ] My code follows the code style of this project.
- [ ] I have read the [CONTRIBUTING](https://github.com/histolab/histolab/blob/main/CONTRIBUTING.md) document.
- [ ] I have tested the changes and verified that they work and don't break anything (as well as I can manage).


.github/workflows/tests.yml
```diff
@@ -50,63 +50,35 @@ jobs:
...
-            sudo apt update
-            sudo apt install openslide-tools
-          elif [ "$RUNNER_OS" == "macOS" ]; then
-            brew install openslide
-          else
-            echo "$RUNNER_OS not supported"
-            exit 1
-          fi
-        shell: bash
-
```
ernestoarbitrio, line 103: ❤️

ernestoarbitrio (approved):

